PR #3, opened 2022-05-21
Add clade

chaoran-chen:
Is it okay to add a column?

Please check for correctness! I only know which sequences are for sure WA and assumed that all the others are CA.


emmahodcroft:
I think `CA` & `WA` is a good way to tackle potentially hairy geographic naming until someone may suggest something else. We don't have any other names to work with yet, but this minimizes the geographic nature!

PR #4, opened 2022-05-22
Add region column

chaoran-chen:


emmahodcroft (approved):

emmahodcroft:
I think this looks good. I suppose a separate question may be whether we can fill in country and region for a few sequences from the name alone?

rneher:
I'll quickly transfer those to the file that has the clade and the new genomes.

PR #7, opened 2022-05-23
Update metadata.tsv

babarlelephant:
In metadata.tsv,
adding a few collection dates of old sequences according to https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4411695/
and changing a few CA clade according to my tree https://nextstrain.org/community/babarlelephant/cov/monkeypoxvirus

Old sequences are important to understand the clockrate and if the 2018->2022 branch really is a big outlier

Thanks a lot for your great work!


rneher:
awesome, thanks!

PR #8, opened 2022-05-23
Simplify installation and build invocation; add CI

tsibley:
Uses an alternate runtime image with Nextalign v2 already baked in.

Also simplified the invocation so the default config file is used by default.

Also added CI to help ensure the build works in the face of changes (from pushes and PRs).

### Related issue(s)
Related to https://github.com/nextstrain/docker-base/pull/43 and https://github.com/nextstrain/.github/pull/23.

### Testing
- [x] `nextstrain build` command shown works for me.
- [x] CI [works](https://github.com/nextstrain/monkeypox/runs/6563968947?check_suite_focus=true) ([output artifacts](https://github.com/nextstrain/monkeypox/suites/6628066966/artifacts/249833482))

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,14 +13,10 @@ Add any additional sequences and metadata in separate fasta or metadata-tsv file
 
 Run pipeline with:
 ```
-snakemake -j 1 -p --configfile=config/config.yaml
+nextstrain build --image=nextstrain/base:branch-nextalign-v2 . --configfile=config/config.yaml
 ```
 
 View results with:
-```
-auspice view --datasetDir auspice/
-```
-or with:
 ```
 nextstrain view auspice/
 ```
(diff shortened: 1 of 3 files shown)


victorlin (approved):
Looks good with the update to `@master` once nextstrain/.github#23 is merged.

README.md
```diff
@@ -92,57 +88,5 @@ uncertain.
...
-  --output-dir=tmp/ \
-  --output-basename=nextalign
-```
-
-Copy `nextalign` binary to somewhere globally accessible
-```
-sudo cp target/release/nextalign /usr/local/bin/
-```
+Follow the [standard installation instructions](https://docs.nextstrain.org/en/latest/install.html) for Nextstrain's suite of software tools.
+Please choose the installation method for your operating system which uses Docker, as currently a pre-release version of Nextalign is required which we've baked into the `--image` argument to `nextstrain build` above.
```
corneliusroemer, line 92: Maybe we should add something more explicit to `nextstrain build`. I just get the following which doesn't tell me much about what the problem is:
```
Warning: The specified --image=nextstrain/base:branch-nextalign-v2 option is not used by --native.
```

Just running the command, but probably missing docker. Warning should tell me what the problem is.

corneliusroemer: The warning should be:
```
Falling back to `--native` since `--docker` does not seem to be available.

Check your setup by running `nextstrain check-setup --set-default`

tsibley: Followup to these comments [happened in Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1653429601942399?thread_ts=1653336973.215419&cid=C01LCTT7JNN), where the comments were cross-posted.

PR #9, opened 2022-05-24
Add metadata

chaoran-chen:


emmahodcroft:
Thanks Chaoran! Adding the accessions is great. I don't know if the plan is to include the new Swiss sequence test, I think @rneher had some questions about some mutations. Of course, including in metadata without a sequence means it won't go into a run, just depends if we want to keep it in sync.

PR #10, opened 2022-05-24
README: Add --docker and --cpus 1 to the example `nextstrain build` invocation

tsibley:
The --image argument is ignored unless a containerized runtime is used
(Docker or AWS Batch currently).  Under the "native"/ambient runtime,
which may be the default for some installations, a warning will be
emitted but ultimately --image will be ignored.  Specifying --docker
explicitly means a more obvious (although still not good) error occurs
earlier.  These two arguments are temporary until Nextalign v2 has a
stable release.

Relatedly, specifying --cpus 1 now heads off issues with
"native"/ambient environments with newer Snakemake's until Nextstrain
CLI accounts for that automatically.  While this invocation doesn't use
the "native"/ambient runtime now, including --cpus now means that when
--image and --docker are no longer necessary we can simply remove them.
Before that point, it also provides a guidance point for people using
variants of this example invocation.

Ultimately, given further work within our ecosystem (Nextalign,
Nextstrain CLI), this whole invocation should revert back to just
`nextstrain build .`.

### Related issue(s)
Related to #8 

### Testing
Ran new example invocation under environments where the
- [x] Docker runtime is default
- [x] Native ("ambient') runtime is default, but Docker runtime is supported
- [x] Docker runtime is not supported

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Add any additional sequences and metadata in separate fasta or metadata-tsv file
 
 Run pipeline with:
 ```
-nextstrain build --image=nextstrain/base:branch-nextalign-v2 .
+nextstrain build --docker --image=nextstrain/base:branch-nextalign-v2 --cpus 1 .
 ```
 
 View results with:
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ Run pipeline with:
 nextstrain build --docker --image=nextstrain/base:branch-nextalign-v2 --cpus 1 .
 ```
 
+Adjust the number of CPUs to what your machine has available you want to perform alignment and tree building a bit faster.
+
 View results with:
 ```
 nextstrain view auspice/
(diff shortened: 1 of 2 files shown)


tsibley:
New commit (1088ed18722ab423ab9b313ac08c0820c956873c) on top:

> Enable parallelism for Nextalign and tree building
>
> Use all the CPUs available for Nextalign, under the assumptions that a) this workflow currently produces a single build at a time and b) Nextalign doesn't have significant upper limits on useful parallelism (number of input sequences not withstanding, which is awkward to calculate here).  These assumptions might be incorrect?  They certainly might _become_ incorrect later, so we should remember to reconsider them in the future.
>
> Use up to 8 CPUs for `augur tree`, as I dimly recall diminishing returns from IQ-TREE above a number like 8.  This might also be wrong, though!

Thoughts?

PR #12, opened 2022-05-25
Add LAPIS entry point

jameshadfield:
While this is slightly circular, it's useful for quick testing and may
become a more standard entrypoint as the ingest becomes automated. The
LAPIS metadata is missing some fields compared to the example metadata
(e.g. "outbreak_associated") but contains some fields the example
metadata does not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,14 +67,23 @@ has been saved to `example_data/outbreak.fasta`.
 
 ### Data preparation
 
-Move metadata to `data/`:
+**Option 1:**
+
+Collect data as described above and store in one or more `data/*.fasta` and `data/*.tsv` file(s).
+
+**Option 2:**
+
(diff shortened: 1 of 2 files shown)


chaoran-chen:
Nice :) Quick comments about the different fields: 

The missing `outbreak_associated` field: I had the impression that the value can be easily derived from the date and thus omitted it. 

The additional fields: They are mostly leftovers from the covid version. I'll get rid of them soon.

rneher:
In this PR https://github.com/nextstrain/monkeypox/pull/14 I made the lapis call produce `data/*` and treat `data/*` as source of truth for other rules. data source can be specified in the config. Ingest could be just another data source that also produces `data/*`

jameshadfield:
Closing in lieu of #14 (which includes this work)

PR #14, opened 2022-05-25
Lapis2

rneher:
This builds on PR #12 and makes the datasource configurable.


rneher:
@jameshadfield does this makes sense to you as a superset of your original PR?

jameshadfield (approved):
Looks great. I think that starting from a single sequences + metadata file (as opposed to `*.fasta`) is a good move too.

PR #19, opened 2022-05-26
metadata for a few 2003 sequences that are not in the nextstrain tree

babarlelephant:
Also includes the metadata and lat_longs and (unknown collection date) correction for the new French and Netherlands sequence

The USA 2003 sequences include the only squirrel in the tree, so worth to add it !


jameshadfield (approved):
Checked the individual GenBank accessions and all look correct. Can't comment on `NL001_2022`. Thanks @babarlelephant!

rneher:
I'll merge this. But one question @babarlelephant: did you add the lat-longs because they were missing? Augur should have most of these countries already covered:
https://github.com/nextstrain/augur/blob/master/augur/data/lat_longs.tsv

PR #27, opened 2022-06-03
Adding exclusions

theosanderson:
This adds one sequence that seems to be missing 3' mutations to both exclude_overview and exclude_outbreak:

```
MPXV_GSTT_Patient1 # reversions in 3' terminal region
```

It additionally copies ~~2~~ the sequence above and an additional one below from exclude_outbreak to exclude_overview and exclude.txt:
```
MPXV_1_IT_Milan_2022  # many extra mutations
```

### DIFF
--- a/config/exclude.txt
+++ b/config/exclude.txt
@@ -51,4 +51,3 @@ W_Nigeria
 
 # potentially problematic
 MPXV_1_IT_Milan_2022  # many extra mutations
-MPXV_FR_HCL0001_2022  # any reversions in terminal regions
--- a/config/exclude.txt
+++ b/config/exclude.txt
@@ -51,3 +51,4 @@ W_Nigeria
 
 # potentially problematic
 MPXV_1_IT_Milan_2022  # many extra mutations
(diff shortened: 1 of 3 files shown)


babarlelephant:
Hi MPXV_FR_HCL0001_2022  has been updated and is fine now

MPXV_GSTT_Patient1 has a lot of deletions and most of its "mutations" are due to misalignment (minimizing gaps) but yes should be excluded

theosanderson:
Thanks @babarlelephant I updated to remove MPXV_FR_HCL0001_2022

theosanderson:
(unnecessary obvious point: it feels to me like it would be easier to maintain these if a single file was used for exclude.txt and exclude_outbreak.txt)

rneher:
thanks. I removed MPXV_FR_HCL0001_2022 from the exclusions. We also moved to organizing exclusion by accession. the exclusion by name files are obsolete. Will clean this up, and move to a single exclusion file soon (once we have clades in the data, we can filter based on those and only need to exclude low quality sequences)

PR #41, opened 2022-06-08
ingest/transform: add `apply-geolocation-rules`

### LINKED ISSUE #34: ingest: adopt geolocation rules
### Context
Use standard geolocation rules to annotate geolocations so that we do not have to make an annotation for the same geolocation edits for multiple records. This would a similar process to how the ncov-ingest uses the [gisaid_geoLocationRules.tsv](https://github.com/nextstrain/ncov-ingest/blob/master/source-data/gisaid_geoLocationRules.tsv).

### Description
Ideally, this would use a centralized geolocation rules TSV (could be within `augur/augur/data/`) for the most general rules.
Then, there can be monkeypox data specific TSV within the repo. 

Within the ingest pipeline, we can fetch the general rules from augur's master branch and concatenate the local rules.
For the function that loads the geolocation rules, we can make sure that the local monkeypox rules can overwrite the general rules. Then include a transform step that overwrites geolocation fields using the full geolocation rules.

joverlee521:
Applies geolocation rules provided in the geolocation rules TSV file to
the geolocation fields (region, country, division, location).
This allows us to create general rules for geolocations that are not
tied to specific record accession (i.e. not user annotations).

Currently fetches the geolocation rules TSV from ncov-ingest, but in the
future this can point to a more centralized location for general rules.

Monkeypox data specific rules in ingest/source-data/geolocation-rules.tsv
will be concatenated to the general rules. Since these local rules are
added at the end, they can overwrite existing general rules.

Closes #34

### DIFF
--- a/ingest/bin/apply-geolocation-rules
+++ b/ingest/bin/apply-geolocation-rules
@@ -1,0 +1,226 @@
+#!/usr/bin/env python3
+"""
+Applies user curated geolocation rules to the geolocation fields in the NDJSON
+records from stdin. The modified records are output to stdout. This does not do
+any additional transformations on top of the user curations.
+"""
+import argparse
+import json
+from collections import defaultdict
+from sys import exit, stderr, stdin, stdout
+
+
+class CyclicGeolocationRulesError(Exception):
+    pass
+
+
+def load_geolocation_rules(geolocation_rules_file):
+    """
+    Loads the geolocation rules from the provided *geolocation_rules_file*.
+    Returns the rules as a dict:
+    {
+        regions: {
+            countries: {
+                divisions: {
+                    locations: corrected_geolocations_tuple
+                }
+            }
+        }
+    }
+    """
+    geolocation_rules = defaultdict(lambda: defaultdict(lambda: defaultdict(dict)))
+    with open(geolocation_rules_file, 'r') as rules_fh:
+        for line in rules_fh:
+            # ignore comments
+            if line.lstrip()[0] == '#':
+                continue
+
+            row = line.strip('\n').split('\t')
+            # Skip lines that cannot be split into raw and annotated geolocations
+            if len(row) != 2:
+                print(
+                    f"WARNING: Could not decode geolocation rule {line!r}.",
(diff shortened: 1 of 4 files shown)


MoiraZuber:
This looks very good to me!

ingest/bin/apply-geolocation-rules
```diff
@@ -0,0 +1,226 @@
...
+        for index, field_value in reversed(list(enumerate(rule_traversal))):
+            if field_value == '*':
+                first_consecutive_general_rule_index = index
+            else:
+                break
+
+        # Edit the raw value of the field directly before the first consecutive '*'
+        # in the rule traversal in hopes that by moving to a general rule on
+        # a higher level, we can find a matching rule.
+        rule_traversal[first_consecutive_general_rule_index - 1] = '*'
```
victorlin, line 133: @joverlee521 I'm a bit confused here, could you provide an example of what this is trying to do?

joverlee521: Yeah, this might be hard to explain. I'll walk through an example.

Example rules:

|rule # | raw | annotated|
|-------------|--------|---------|
| 1 | a/b/c/d | 1/2/3/4 |
| 2 | a/\*/\*/\* | 1/\*/\*/\* |

A example case: `( a, b, c, e )`

Here is how `rule_traversal` will be edited per functional call:
1. [a]
2. [a, b]
3. [a, b, c] <- Up to this point, we are matching rule # 1
4. [a, b, c, e] <- This doesn't match any available rules, so we try to backfill with `*`
5. [a, b, c, * ] <- Still no matching rule, continue `*` backfill
6. [a, b, *, *] <- Still no matching rule, continue `*` backfill
7. [a, *, *, * ] <- This now matches rule # 2!

And of course as I explain this, I'm realizing there are cases where this logic would not work...I'll create a PR to update this tomorrow

joverlee521: See https://github.com/nextstrain/monkeypox/pull/55 for the updated logic 🙏

PR #42, opened 2022-06-08
Add --cpus option

pvanheus:
Based on my experience with latest Nextstrain, `--cpus` option is required so this PR adds those in the relevant places.

Make a minor fix to the documentation for the ingest step.

### Testing
Tested by running the code suggested command line.

### DIFF
--- a/ingest/README.md
+++ b/ingest/README.md
@@ -11,13 +11,13 @@ This is the ingest pipeline for Monkeypox virus sequences.
 Fetch sequences with
 
 ```sh
-nextstrain build . data/sequences.ndjson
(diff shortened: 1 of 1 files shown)


joverlee521 (approved):
Thank you @pvanheus!

tsibley:
Note that it's only required (in a sense) when using the "native" runtime with Snakemake 5.11 or later. The Docker runtime does not require it. Relatedly, there's [this proposed change](https://github.com/nextstrain/cli/pull/179) which will make it optional regardless of runtime or Snakemake version.

PR #43, opened 2022-06-08
Add `institution` to MPX metadata

emmahodcroft:

Adds manually-curated institution information for most WA clade MPX seqs
- Adds data to `souce-data/annotations.tsv`
- Adds an extra field `institution` to the final metadata file (we may want to amend this field name)

Will not have info for any sequences added since Saturday.

We likely want to try to automate this in future from INSDC data, but this makes a good start at better crediting sequence submitters.

The edit to `merge-user-metadata` is because while testing I noticed that multiple empty lines at the end of `annotations.tsv` or any empty line in the middle of the file caused an error. I think it's fine to handle this more robustly and just ignore empty lines.

### Testing
Ran locally, seemed to output appropriately, but extra eyes appreciated.

### DIFF
--- a/ingest/bin/merge-user-metadata
+++ b/ingest/bin/merge-user-metadata
@@ -33,7 +33,7 @@ if __name__ == '__main__':
     with open(args.annotations, 'r') as annotations_fh:
         csv_reader = csv.reader(annotations_fh, delimiter='\t')
         for row in csv_reader:
-            if row[0].lstrip()[0] == '#':
+            if not row or row[0].lstrip()[0] == '#':
                     continue
             elif len(row) != 3:
                 print("WARNING: Could not decode annotation line " + "\t".join(row), file=stderr)
--- a/ingest/config/config.yaml
+++ b/ingest/config/config.yaml
@@ -58,4 +58,5 @@ transform:
     'clade',
     'abbr_authors',
     'authors',
+    'institution'
   ]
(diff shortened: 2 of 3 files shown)


joverlee521:
I manually started a [test run](https://github.com/nextstrain/monkeypox/actions/runs/2464110500) for this branch via GitHub Actions UI. The files will be uploaded to `s3://nextstrain-data/files/workflows/monekypox/branch/add_institution`.

joverlee521 (approved):
Thanks for curating the institutions @emmahodcroft! The metadata file from the test run (s3://nextstrain-data/files/workflows/monkeypox/branch/add_institution/metadata.tsv.gz) looks good to me.

I have a single question about whether we want to use the column name `institution`. Do we want to match ncov by labeling the column as `originating_lab` or `submitting_lab`? (Not entirely sure which field institution aligns with...)

I'm hoping we have a "standard" set of columns for all pathogen metadata so that users don't have to struggle with the metadata format for each pathogen. Granted, there are definitely columns specific to each pathogen...

emmahodcroft:
There was a little chat about this on Slack (mostly me and Chaoran) and Chaoran suggested `institution` over my original `lab`. It would be nice to have originating & submitting, but as far as I can tell from Genbank, there is usually only one lab? If there were two, I'm not sure if we'd be able to tell which was which?

In general I think most labs just want to be recognised/credited, and us trying to sort out which is which (unless Genbank actually does have a way to distinguish this already, that I totally missed) is probably going to be impossible. However, we may want to be able to include more than 1 lab. In that case, we could switch this column to be `institutions` and designate some kind of delimiter so that they could be split up for display purposes. This may be more robust long-term as people could in theory include as many labs as they feel they should credit (processing_lab, leading_lab, etc).

PR #45, opened 2022-06-09
Robust transform-date-fields

joverlee521:
My attempt to make the ingest `transform-date-fields` more robust. 
See commits for details. 

---
Manually ran [ingest on branch](https://github.com/nextstrain/monkeypox/actions/runs/2465474800), resulting files will be available at `s3://nextstrain-data/files/workflows/monkeypox/branch/robust-format-date/`

### DIFF
--- a/ingest/bin/transform-date-fields
+++ b/ingest/bin/transform-date-fields
@@ -9,7 +9,7 @@ from datetime import datetime
 from sys import stderr, stdin, stdout
 
 
-def format_date(date_string: str, expected_formats: set) -> str:
+def format_date(date_string: str, expected_formats: list) -> str:
     """
     Originally from nextstrain/ncov-ingest
 
@@ -19,7 +19,7 @@ def format_date(date_string: str, expected_formats: set) -> str:
     If *date_string* is an incomplete date (i.e. missing month or day), then
     missing values are masked with 'XX'.
 
-    >>> expected_formats = {'%Y-%m-%d', '%Y-%m-%dT%H:%M:%SZ', '%m-%d'}
+    >>> expected_formats = ['%Y-%m-%d', '%Y-%m-%dT%H:%M:%SZ', '%m-%d']
 
     >>> format_date("01-01", expected_formats)
     'XXXX-XX-XX'
@@ -135,11 +135,12 @@ if __name__ == '__main__':
     parser.add_argument("--date-fields", nargs="+",
         help="List of date field names in the NDJSON record that need to be standardized.")
     parser.add_argument("--expected-date-formats", nargs="+",
-        help="Expected date formats that are currently in the provided date fields")
+        help="Expected date formats that are currently in the provided date fields." +
+             "If a date string matches multiple formats, it will be parsed as the first format in the list.")
 
     args = parser.parse_args()
 
-    expected_formats = set(args.expected_date_formats)
+    expected_formats = args.expected_date_formats
 
     for record in stdin:
         record = json.loads(record)
--- a/ingest/bin/transform-date-fields
+++ b/ingest/bin/transform-date-fields
@@ -15,8 +15,14 @@ def format_date(date_string: str, expected_formats: set) -> str:
 
     Format *date_string* to ISO 8601 date (YYYY-MM-DD).
     If *date_string* does not match *expected_formats*, return *date_string*.
+    If *date_string* is missing the year, return masked date 'XXXX-XX-XX'.
+    If *date_string* is an incomplete date (i.e. missing month or day), then
+    missing values are masked with 'XX'.
 
-    >>> expected_formats = {'%Y-%m-%d', '%Y-%m-%dT%H:%M:%SZ'}
+    >>> expected_formats = {'%Y-%m-%d', '%Y-%m-%dT%H:%M:%SZ', '%m-%d'}
+
+    >>> format_date("01-01", expected_formats)
+    'XXXX-XX-XX'
 
     >>> format_date("2020", expected_formats)
     '2020-XX-XX'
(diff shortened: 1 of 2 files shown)


emmahodcroft:
The code is beyond my reviewing capability, but I really like the ideas here - seems like a really robust solution Jover! This would be a great thing to integrate into `auspice` as part of (or attached to) `transform`, too. This is such a common issue.

ingest/bin/transform-date-fields
```diff
@@ -36,14 +42,79 @@ def format_date(date_string: str, expected_formats: set) -> str:
...
+    #
+    # Allows us to check if year/month/day are included in the date format so we
+    # know when to mask incomplete dates with 'XX'
+    all_field_directives = {'%c', '%x',
+        ('%G', '%V', '%A'), ('%G', '%V', '%a'), ('%G', '%V', '%w'), ('%G', '%V', '%u')
+    }
+    month_and_day_directives = {'%j',
+        ('%U', '%A'), ('%U', '%a'), ('%U', '%w'), ('%U', '%u'),
+        ('%W', '%A'), ('%W', '%a'), ('%W', '%w'), ('%W', '%u')
+    }
```
tsibley, line 56: Hmm, I'm not sure it's valid to combine ISO week-based numbering (`%G`, `%u`, `%V`) with non-ISO week-based numbering (`%U`, `%W`, `%a`, `%A`, `%w`), as the two systems are "indexed" differently. So we may want to drop `('%U', '%u')` and `('%W', '%u')` from `month_and_day_directives` and `('%G', '%V', '%A'), ('%G', '%V', '%a'), ('%G', '%V', '%w')` from `all_field_directives`.

ingest/bin/transform-date-fields
```diff
@@ -36,14 +42,79 @@ def format_date(date_string: str, expected_formats: set) -> str:
...
+        Checks if any of the directives in *potential_directives* is included
+        in *date_format* string.
+
+        If an element within *potential_directives* is a tuple, then all directives
+        within the tuple must be included in *date_format*.
+        """
+        return any(
+            (
+                (isinstance(directive, str) and directive in date_format) or
+                (isinstance(directive, tuple) and all(sub_directive in date_format for sub_directive in directive))
```
tsibley, line 72: While the `in` operator is super nice for readability here, it doesn't handle the edge case of a format string containing an escaped directive (e.g. `%%Y` means literal `%Y` not "a 4+ digit year").

Regexps provide a nice way to handle this with negative lookbehind assertions, e.g. `(?<!%)%Y`, generalized to something like:

```python
def included(directive, format):
    return bool(re.search(f"(?<!%){re.escape(directive)}", format))
```

ingest/bin/transform-date-fields
```diff
@@ -36,14 +42,79 @@ def format_date(date_string: str, expected_formats: set) -> str:
...
+                day_string = parsed_day_string
+
+            # If not directives for BOTH month and day are included, then check
+            # month directive was included in date format
+            elif(directive_is_included(month_directives, date_format)):
+                month_string = parsed_month_string
+
+                # Only check for day directives if month is included
+                if(directive_is_included(day_directives, date_format)):
+                    day_string = parsed_day_string
```
tsibley, line 116: This cascade of checks to greater specificity makes sense as an implementation approach. It also serves an important functional role because some directives are dependent on others to parse correctly, e.g. `%U` requires `%Y` (or `%y`), and we enforce that thru this cascade.

However, I wonder if the sets of directives above and this code would be easier to reason about if we explicitly listed the combinations for each level instead of cascading. It'd be more verbose, but I think not overly so.

For example, what if instead of cascading down these sets

```
all_field_directives
month_and_day_directives
year_directives
month_directives
day_directives
```

we used a flat if/elif/elif chain against these sets:

```
year_month_day_directives
year_month_directives
year_directives
```

ingest/bin/transform-date-fields
joverlee521: Hmm, I included them since they were valid combinations for `datetime.strptime` based on the error message when trying to parse an incomplete ISO directive: 
```
ValueError: ISO year directive '%G' must be used with the ISO week directive '%V' and a weekday directive ('%A', '%a', '%w', or '%u').
```

I did see that the Year/week directives cannot be interchanged between ISO/non-ISO, I think that matches what you mean by "indexed" differently? I would think the weekday directives can be used across ISO/non-ISO directives.

ingest/bin/transform-date-fields
joverlee521: I think the current way makes it easier to add new directives. However, that may not be a worry since the directives rarely get updated 🤔

tsibley: Yeah, the directives are likely to change approximately never. ;-) And in general, I think the default should be to optimize the code for reading, not writing, as code will be read many many more times by many more people than it is written.

(But this doesn't mean you have to take my suggestion here!)

tsibley:
Hmm. Realizing that there are also probably platform-specific directives that we should include here, as they can be successfully used if someone's only running `augur` on platforms that support more than what C89 does. The Python docs mention this:

> The full set of format codes supported varies across platforms, because Python calls the platform C library’s strftime() function, and platform variations are common. To see the full set of format codes supported on your platform, consult the [strftime(3)](https://manpages.debian.org/strftime(3)) documentation. There are also differences between platforms in handling of unsupported format specifiers.

ingest/bin/transform-date-fields
tsibley: Ah, hmm. I guess they're technically valid, in the sense that datetimes can be formatted with them and re-parsed back successfully. Ok to keep them then.

I guess I was more thinking about if they're valid for reasonable use, as you can't tell unambiguously from just the formatted datetime how it should be parsed, so it's a bit "dangerous" to mix them. For example, `2022 01 Sun` (year, week, weekday) means different days depending on how it was produced:

```
>>> datetime.strptime("2022 01 Sun", "%Y %U %a")
datetime.datetime(2022, 1, 2, 0, 0)
>>> datetime.strptime("2022 01 Sun", "%G %V %a")
datetime.datetime(2022, 1, 9, 0, 0)
```

By "indexed" differently, I mean the starting points for counting weeks in a year and the weekday of weeks.

joverlee521: Yeah, I guess we would be leaving it up to the user to understand the datetime directives and provide the correct combinations. The same problem exists even within non-ISO:
```
>>> datetime.strptime("2022 01 Sun", "%Y %U %a")
datetime.datetime(2022, 1, 2, 0, 0)
>>> datetime.strptime("2022 01 Sun", "%Y %W %a")
datetime.datetime(2022, 1, 9, 0, 0)
```

ingest/bin/transform-date-fields
joverlee521: Yeah, I might not put in the effort here in monkeypox since I expect it to be replaced by the new augur command, but will definitely go for readability in the augur version of date transformations.

ingest/bin/transform-date-fields
tsibley: Yep! Ok to leave as is.

victorlin:
This is neat! Much broader support than [Augur's dates.py](https://github.com/nextstrain/augur/blob/bf400de475d61af188acc6752e265cfd6469e834/augur/dates.py), but also noting that incomplete dates such as `2020`, `2020-01` which are addressed here, are also addressed in current Augur date support (though it isn't perfect: nextstrain/augur#747).

Once this makes its way into the Augur subcommand, we can probably scope downstream Augur date support to just:

1. Exact ISO-8601 dates
2. Ambiguous ISO-8601-formatted dates
3. Numeric dates
4. Relative dates

Related: nextstrain/augur#928

PR #47, opened 2022-06-09
Force line endings to be LF on check out

huddlej:

Fixes an issue for users running workflows on Windows where git clone will convert line endings to CRLF by default, even though users will run the workflows in a UNIX-style system through Docker. This commit forces line endings to be LF regardless of the platform.

### Testing

 - [ ] Tested with Mac OS
 - [ ] Tested with Windows

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Always use LF line endings even on Windows.
+* text eol=lf


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Always use LF line endings even on Windows.
+* text eol=lf
```
tsibley, line 2: I don't think we want to force the `text` attribute here. While we don't yet have any non-text (binary) files in the repo at the moment, we could easily have them in the future (`.gz`, `.png`, etc.) and forcing text mode will corrupt them.

I think it should be enough to set `eol=lf` and let git automatically detect text files per it's normal config? Or, if we want to force automatic detection regardless of config, we could specify `text=auto`.

`man gitattributes` says:

```
Set
   Setting the text attribute on a path enables end-of-line
   normalization and marks the path as a text file. End-of-line
   conversion takes place without guessing the content type.

Unset
   Unsetting the text attribute on a path tells Git not to attempt
   any end-of-line conversion upon checkin or checkout.

Set to string value "auto"
   When text is set to "auto", the path is marked for automatic
   end-of-line conversion. If Git decides that the content is
   text, its line endings are converted to LF on checkin. When the
   file has been committed with CRLF, no conversion is done.

Unspecified
   If the text attribute is unspecified, Git uses the
   core.autocrlf configuration variable to determine if the file
   should be converted.
```

PR #52, opened 2022-06-13
CloudFront Invalidation

joverlee521:
Invalidate CloudFront after file has been uploaded to S3 so that we serve latest file at data.nextstrain.org.

(Adding the same CloudFront invalidation in [ncov-ingest](https://github.com/nextstrain/ncov-ingest/pull/315))

### DIFF
--- a/ingest/README.md
+++ b/ingest/README.md
@@ -56,6 +56,25 @@ Do the following to include sequences from static FASTA files.
 Configuration takes place in `config/config.yaml` by default.
 Optional configs for uploading files and Slack notifications are in `config/optional.yaml`.
 
+### Environment Variables
+
+The complete ingest pipeline with AWS S3 uploads and Slack notifications uses the following environment variables:
+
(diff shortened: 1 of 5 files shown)


joverlee521:
Thank you for your feedback @trvrb! I've added the environment variables to the ingest/README in ea63c7f.

I did the following for testing:
1.  I manually uploaded a metadata.tsv file with only two records to `s3://nextstrain-data/files/workflows/monkeypox/branch/cloudfront-invaldiation/metadata.tsv.gz`
2. Then I ran the [cloudfront-valdiation branch GH action workflow](https://github.com/nextstrain/monkeypox/actions/runs/2498193427). 
3. When the pipeline finished, I immediate pulled the metadata TSV from https://data.nextstrain.org/files/files/workflows/monkeypox/branch/cloudfront-invaldiation/metadata.tsv.gz and verified that there metadata file was updated with the latest records.

PR #54, opened 2022-06-13
ingest: Add tests

victorlin:

I wrote these tests so I can understand how things work, but they also might be good to have in the repo. Filenames and tests location up for debate.

If this is desired, we don't have to add tests for all Python scripts in this PR, but it would be good to rename the scripts with underscores and `.py` extension here.

### Related issue(s)

_N/A_

### Testing

See CI.


ingest/bin/tests/test_apply_geolocation_rules.py
```diff
@@ -89,3 +90,45 @@ def test_get_annotated_geolocation_wildcards(self, tmpdir):
...
+        assert get_annotated_geolocation(geolocation_rules, ("a", "b", "c", "d")) == ("1", "2", "3", "4")
+
+    def test_wildcards_work_if_partial_match_exists(self, tmpdir):
+        """Test wildcards still work even when partial matches exist."""
+        lines = (
+            "a/b/c/d\t1/2/3/4",
+            "a/*/*/z\t1/*/*/26",
+        )
+        geolocation_rules = get_geolocation_rules(tmpdir, lines)
+        assert get_annotated_geolocation(geolocation_rules, ("a", "b", "c", "z")) == ("1", "*", "*", "26")
```
joverlee521, line 134: This test fails on master, but passes if rebased onto #55

victorlin:
This only adds tests for `apply_geolocation_rules`, which has moved to nextstrain/ingest. Closing in favor of adding the tests in that repo: https://github.com/nextstrain/ingest/pull/26

PR #57, opened 2022-06-15
mpxv temporal colouring scale

jameshadfield:
Note: this requires forthcoming PRs in augur and auspice, and I'll notify here when released etc. Because of this i'll keep this as a draft PR.

The end result is 
image


jameshadfield:
Augur PR: https://github.com/nextstrain/augur/pull/969
Auspice PR: https://github.com/nextstrain/auspice/pull/1523

jameshadfield:
Both of the prereq PRs above have been merged. Auspice is released (2.37.3) and the augur change should be released tomorrow 🤞

victorlin:
> Wait until the nextstrain image has been updated (should be automatic)

This has been done: [`nextstrain/base:build-20220616T211939Z`](https://hub.docker.com/layers/base/nextstrain/base/build-20220616T211939Z/images/sha256-0a5510fc481194b62dec38a38f417f5313376db46469147c172dc8b1019206c4)

jameshadfield:
Added a commit to ensure we're running augur v16, which is causing CI to fail as `nextstrain/base:branch-nextalign-v2` has yet to be updated...

corneliusroemer:
I've restarted failing CI to see if the new image has propagated by now @jameshadfield

corneliusroemer:
I'll merge because it's better than what we use currently. But is it on purpose that there's no cake for the map?
image

jameshadfield:
> but I leave the button pushing to you

Thanks for merging this -- it's the nice thing about working across different time zones 😂 

> But is it on purpose that there's no cake for the map?

I've not head them called 🍰  but I like it. Continuous scales (and temporal is a continuous scale with small tweaks) get colour blending (e.g. https://next.nextstrain.org/zika/?c=num_date&d=map&p=full), however it's been a to-do to explore discretising these (e.g. via the binning in the legend) to produce a 🍰 . It would certainly make the map code simpler!

PR #62, opened 2022-06-16
Remove Finnish Virological sequences from repo

trvrb:
Sequences MPX-37/Finland/2022 and MPX-42/Finland/2022 have been added to GenBank. This commit:
1. removes the versioned `.ndjson` file with these data
2. removes reference to this `.ndjson` from the ingest config
3. add annotations for date for both these sequences as the GenBank records are missing data information
4. fixes sort order for annotations.tsv

This PR supersedes PR #60 (which was missing items 2 and 3 above) as well as PR #61 (which was missing items 1 and 3 above).


joverlee521:
I also removed the NDJSON from `.gitignore` and updated the accessions in `example_data/metadata.tsv` (although it's not clear to me if the example data file is still used?)

trvrb:
I think it's now safe to remove example_data. At the moment I think it's likely to just add confusion (and adds another thing we need to maintain).

victorlin:
`example_data/` is currently being used by the [CI checks](https://github.com/nextstrain/monkeypox/pull/62/checks), which mostly just checks that the workflow runs without error:

image

If this check is valuable, is there some other metadata.tsv/sequences.fasta pair that would be appropriate to swap in here? Or another option, to avoid confusion, could be to mention that `example_data/` is just for the CI checks now.

trvrb:
Good catch. Thank you @victorlin. My suggestion in this case would be to keep `example_data/` but scope it more narrowly so that it's just enough data to run as CI check. This is similar to how https://github.com/nextstrain/zika uses `example_data/` but only includes 34 sequences instead of the live build's 850.

PR #67, opened 2022-06-17
add updated clades

rneher:
replace reference with "official reference" `NC_063383.1`.

@corneliusroemer: this currently uses the locus tags as names for the genes. This is probably not what we want, right?


corneliusroemer:
We should use the GFF3 genemap rather than the `.gb` file. Then it works (using my patched Augur version at least).

It's easier to fix a GFF3 file than rename things in .gb file (not so easily column structured).

Replacing translate like this works for me
```
rule translate:
    message: "Translating amino acid sequences"
    input:
        tree = rules.refine.output.tree,
        node_data = rules.ancestral.output.node_data,
        genemap = config["genemap"] # <---- changed
    output:
        node_data = build_dir + "/{build_name}/aa_muts.json"
    shell:
        """
        augur translate \
            --tree {input.tree} \
            --ancestral-sequences {input.node_data} \
            --reference-sequence {input.genemap} \  #<--- changed
            --output {output.node_data}
        """
```

corneliusroemer:
OK, I need to patch `augur translate` for this to work generally. Since I get this error otherwise:
```
Validating schema of 'results/hmpxv1/raw_tree.json'...
        ERROR: 'nuc' is a required property. Trace: properties - meta - properties - genome_annotations - required
Validation of 'results/hmpxv1/raw_tree.json' failed.
```

corneliusroemer:
This PR switches this branch to genemap from gb file
https://github.com/nextstrain/monkeypox/pull/69

PR #69, opened 2022-06-17
feat: use GFF3 genemap instead of genbank reference so that genenames are OPG

corneliusroemer:
Needs `augur translate` to be patched first before merging. @rneher


jameshadfield (approved):
Tested using https://github.com/nextstrain/augur/pull/976 and all is working 👍

PR #70, opened 2022-06-18
Provision colors separately for mpxv and hmpxv1 build targets

trvrb:
The previous strategy of assigning colors used the entire `metadata.tsv` to assign country colors, whereas the `hpmxv1` build target has a subset of 18 of these 30 countries and the `mpxv` build target has a subset of 26 of these 30 countries. Consequently, we weren't using color spectrum as efficiently as we could especially for the `hmpxv1` target.

This PR updates `filter` rule to output filtered metadata for both `hmpxv1` and `mpxv` and uses this filtered metadata to assign colors.

Additionally, this commit updates country order in `update_colours.py` to start with Africa rather than Asia. For `ncov`, I had started with Asia as this was the basal region. For monkeypox this is Africa.

Out from running this PR can be seen at [nextstrain.org/staging/monkeypox/hmpxv1](https://nextstrain.org/staging/monkeypox/hmpxv1/colors-for-resolution?c=country)

Screen Shot 2022-06-18 at 3 43 56 PM


jameshadfield (approved):
Region-ordering based on proximity to origin makes sense 👍

PR #72, opened 2022-06-18
Include "accession" in metadata

trvrb:
Rather than replacing "accession" with "strain" in the initial `wrangle_metadata.py` script, instead keep "accession" column and create an additional "strain" column. This will allow `export` rule to recognize this "accession" column and include field in tooltips.

Also, remove `genbank_accession_rev` as "Accession" coloring. Now "Accession" will be automatically recognized and displayed in tooltip without having to have a coloring provided.

### DIFF
--- a/config/auspice_config_hmpxv1.json
+++ b/config/auspice_config_hmpxv1.json
@@ -50,11 +50,6 @@
       "key": "date_submitted",
       "title": "Submission Date",
       "type": "categorical"
-    },
-    {
-      "key": "genbank_accession_rev",
-      "title": "Accession",
-      "type": "categorical"
     }
   ],
   "geo_resolutions": [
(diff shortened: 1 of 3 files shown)


jameshadfield (approved):
Code looks good & ran a test build locally to confirm that accession was available in the tip-clicked info panel. Ideally it'd be a clickable link to GenBank, however I'll look into this separately...

joverlee521:
> Ideally it'd be a clickable link to GenBank, however I'll look into this separately...

I did a quick search of Auspice to see how the accession/url works. The special treatment of accessions comes from [`AccessionAndUrl`](https://github.com/nextstrain/auspice/blob/9166cf304fc0d6f4da67bab94a3a63931fd819a1/src/components/tree/infoPanels/click.js#L55), where the accession gets associated with a url if it meets one of these conditions:
1. the node has a `genbank_accession`
2. the node has a `gisaid_epi_isl` -> 
3. the node has both an `accession` and a `url`

I think we want to use condition (3) since we support arbitrary non-GenBank sequences for monkeypox. 
I can added the `url` field in ingest for GenBank sequences. For arbitrary sequences, the `url` would have to be added through manual annotations.

PR #73, opened 2022-06-20
chore: make submission_date scale temporal, parallel to mpxv change

corneliusroemer:
I mirrored the changes @jameshadfield made to mpxv build to hmpxv1 in #57 

I'm not quite sure what the effect is, nothing seems to have changed to my eyes. So maybe this is not supposed to be changed? It doesn't do any harm though, either.

@jameshadfield can you enlighten me what the temporal scale does differently than not specifying it?

### Testing
Ran it locally, looked good

### DIFF
--- a/config/auspice_config_hmpxv1.json
+++ b/config/auspice_config_hmpxv1.json
@@ -12,6 +12,11 @@
   "build_url": "https://github.com/nextstrain/monkeypox",
   "colorings": [
     {
+      "key": "sample_date",
+      "title": "Sample Date",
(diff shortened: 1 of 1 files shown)


corneliusroemer:
I see, in that case I'll close this PR, makes sense!

jameshadfield:
Yeah -- `numdate` (or `num_date`) were the only colorings which could be temporal in auspice, and they had to be defined across all nodes in the tree (e.g. via timetree). My changes open this up to arbitrary colourings.

PR #81, opened 2022-06-26
remove hxmpv1 exclude file, use clade to filter

rneher:
this removes one of the exclude files (`hmpxv1`) and replaces it by the generic one for the larger clade. Filtering is instead done by a `--exclude-where clade!=hmpxv` via a generic `filters` field in the config.


rneher:
this currently fails because the example data doesn't has the clade column

PR #84, opened 2022-06-28
chore: format using snakefmt

corneliusroemer:
### Upstream

I have noticed two issues with snakefmt:
1. Too many newlines between `include:` and `configfile:` statements, especially when they are part of `if/else` blocks: https://github.com/snakemake/snakefmt/issues/149
2. Ternary operator is formatted in a way that reduces readability: https://github.com/snakemake/snakefmt/issues/150

I've created issues in snakefmt, in the hope that this can be addressed.

If you have other issues, please add them here, it'd be great if we could start using snakefmt at some point.

Applies `snakefmt` formatting to Snakemake files

Install through:
```bash
mamba install -c bioconda snakefmt
```

Run using
```bash
snakefmt .
```
in the repo root before committing.

### Testing
I ran the build and it worked.


victorlin (changes requested):
If we want to adopt this, would be great to check this in CI like so: 28fcedcbeca7a8a0064a62d204c03abc39a406f8 ([see example run](https://github.com/victorlin/monkeypox/runs/7101580244?check_suite_focus=true))

corneliusroemer:
Agreed @victorlin, one could also use a precommit hook - though I've never used one.

Also, while we're at it, we may want to start using a Python formatter - maybe black? It's good to trial it with monkeypox as it's a nice and small repo, we can then extend to other repos later once we've got a bit more experience.

.github/workflows/ci.yaml
```diff
@@ -5,8 +5,19 @@ on:
...
+
+  lint:
+     runs-on: ubuntu-latest
+     steps:
+     - uses: actions/checkout@v3
+     - uses: actions/setup-python@v4
+       with:
+         python-version: ~3
+     - run: pip install snakefmt
+     - run: snakefmt --check .
```
victorlin, line 22: I forgot to add in my original comment - along with the CI check, it would be equally appropriate to add developer documentation on how to check for and fix lint issues. Maybe create a `CONTRIBUTING.md` or `docs/DEV_DOCS.md` starting with this:

---

# Developer guide

## Snakemake formatting

We use [snakefmt](https://github.com/snakemake/snakefmt) to ensure consistency in style across Snakemake files in this project.

1. [Install snakefmt](https://github.com/snakemake/snakefmt#install)
2. Check for styling issues with `snakefmt --check .`
3. Automatically fix styling issues with `snakefmt .`

victorlin (changes requested):

.github/workflows/ci.yaml
corneliusroemer: Done! Also added `recommended workspace extensions` for VScode, making it easy to install snakefmt without searching for it.

.vscode/extensions.json
```diff
@@ -0,0 +1,7 @@
+{
```
victorlin, line 1: Unless a substantial amount of contributors use VSCode, I'm a bit wary of adding this file to the repo – it's another thing to maintain even if infrequently used, and will be "noise" to non-VSCode users.

workflow/snakemake_rules/CONTRIBUTING.md
```diff
@@ -0,0 +1,30 @@
...
+### IDE-independent
+
+1. Check for styling issues with `snakefmt --check .`
+1. Automatically fix styling issues with `snakefmt .`
+
+### Using VSCode extension
+
+1. Install the [VSCode extension](https://marketplace.visualstudio.com/items?itemName=tfehlmann.snakefmt)
+1. Check for styling issues with `Ctrl+Shift+P` and select `snakefmt: Check`
+1. Automatically fix styling issues with `Ctrl+Shift+P` and select `Format document`
```
victorlin, line 30: _(non-blocking)_

On the other hand, I think having these VSCode-specific steps is fine since the IDE-independent alternative is also documented.

workflow/snakemake_rules/CONTRIBUTING.md
```diff
@@ -0,0 +1,30 @@
...
+
+```bash
+mamba install -c bioconda snakefmt
+```
+
+- Using pip:
+
+```bash
+pip install snakefmt
+```
```
victorlin, line 19: _(non-blocking)_

My preference is to link out to https://github.com/snakemake/snakefmt#install, but I can see how this is easier for the reader to follow along.

workflow/snakemake_rules/CONTRIBUTING.md
```diff
@@ -0,0 +1,30 @@
+# Developer guide
```
victorlin, line 1: Suggestion: put this file at the top level of the repo, so we can add non-Snakemake specific sections to it later.

victorlin (changes requested):

tsibley:
Linting is great, but FWIW, I'm not a big fan of auto-formatting and would prefer not to require auto-formatting tools. I expect I may be in a minority here, but is there any way to use snakefmt just for linting?

workflow/snakemake_rules/CONTRIBUTING.md
corneliusroemer: Oh this was an accident... Thanks!

corneliusroemer:
The error code for error is `123`, so we could do simpler:
image

The linter could be better - it often doesn't say where the error is - but I guess better than nothing.

I'm also not sure it's possible to lint without formatting using the extension. I'll open an enhancement issue on the repo.

I'd kind of like to try out how helpful the formatting is - we can always go back if it's annoying us. This is a small repo and won't have major reverberations.

.vscode/extensions.json
corneliusroemer: I think almost everyone uses VSCode. Everyone in neherlab uses vscode - and you do too?

But sure, this was just a neat thing that I came across when trying to link to the extension.

victorlin: I do, but I can't speak for others in @blab or @nextstrain as a whole. Just wanted to point this out because I noticed that [no other @nextstrain repos have `.vscode/` checked in](https://cs.github.com/?scopeName=All+repos&scope=&q=org%3Anextstrain+path%3A.vscode) and I haven't noticed it much among other open source projects.

Consider this non-blocking though, unless anyone else has a strong opinion here.

tsibley: I don't use VSCode and tend to think editor configs should live outside the codebase, but I'm also fine with committing it if folks on the team will benefit and it's not high churn.

tsibley:
> @tsibley I'd kind of like to try out how helpful the formatting is - we can always go back if it's annoying us. This is a small repo and won't have major reverberations.

Yep, that's fine by me. I don't want to stand in the way of the people doing the work in this repo. My comment was mostly to add in my 2¢ since it was requested (here and in Slack), esp. given the broader implications in:

> Also, while we're at it, we may want to start using a Python formatter - maybe black? It's good to trial it with monkeypox as it's a nice and small repo, we can then extend to other repos later once we've got a bit more experience.

corneliusroemer:
Thanks @tsibley for your thoughts!

You're right to raise downsides of formatters. 

Let's agree to try snakefmt here, see how it goes, whether it helps catch errors when developing without limiting custom formatting.

We can move back to non-formatted whenever we want.

Regarding linting, unfortunately I don't think this adds much as a workflow with linting will fail the CI snakemake run - unless the CI run doesn't test all.

So maybe it helps in this repo because it would catch ingest which isn't tested in CI. But generally the benefit would be limited.

Still no reason not to add it of course for untested code paths

.vscode/extensions.json
corneliusroemer: I hear you - don't think there will be high churn, this is just to help set up snakefmt for now.

This has 0 impact on most folks since it's just "recommendation".

PR #86, opened 2022-06-28
chore: update example data, generated by chores.smk

corneliusroemer:
Updates example data with the output of chores.smk

### Tests
Failing, oops, we seem to have duplicates in metadata.tsv? @victorlin


victorlin:
[Oh no...](https://github.com/nextstrain/monkeypox/runs/7099066330?check_suite_focus=true)

```
ERROR: Duplicate found in 'results/metadata.tsv'.
```

Seems like there is a duplicate in `accession` now? I had used that column instead of `strain` to avoid duplicates:

https://github.com/nextstrain/monkeypox/blob/8870abdf2ded7051adf57131b7df5d6d791a52fe/workflow/snakemake_rules/chores.smk#L8

We might need an extra step here before `augur filter` to de-duplicate.

corneliusroemer:
I've updated the chores.smk to no longer produce duplicate metadata in #87

PR #87, opened 2022-06-28
fix: duplicate metadata output from chores

corneliusroemer:
Fixes problem that there's duplicate metadata produced by chores.smk, see https://github.com/nextstrain/monkeypox/pull/86#issuecomment-1169098568

### DIFF
--- a/workflow/snakemake_rules/chores.smk
+++ b/workflow/snakemake_rules/chores.smk
@@ -9,13 +9,14 @@ rule update_example_data:
 
     TODO: Use `strain` as the ID column after https://github.com/nextstrain/monkeypox/issues/33 is done.
     """
-    message: "Update example data"
+    message:
+        "Update example data"
     input:
-        sequences = "data/sequences.fasta",
-        metadata = "data/metadata.tsv"
+        sequences="data/sequences.fasta",
+        metadata="data/metadata.tsv",
     output:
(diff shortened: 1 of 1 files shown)


workflow/snakemake_rules/chores.smk
```diff
@@ -28,4 +29,7 @@ rule update_example_data:
             --subsample-seed 0 \
             --output-metadata {output.metadata} \
             --output-sequences {output.sequences}
+        cp {output.metadata} /tmp/metadata-pipe
+        tsv-uniq /tmp/metadata-pipe > {output.metadata}
+        rm /tmp/metadata-pipe
```
victorlin, line 34: Is there a need for `cp`?

```suggestion
            --output-metadata /tmp/metadata-pipe \
            --output-sequences {output.sequences}
        tsv-uniq /tmp/metadata-pipe > {output.metadata}
        rm /tmp/metadata-pipe
```

corneliusroemer: Right, I was trying to do something more complicated and then forgot there was an easy way. You're right! You can push a fix straight to master or we wait until the duplicate bug is fixed upstream

corneliusroemer:
This can be reverted once https://github.com/nextstrain/augur/issues/985 is fixed

PR #92, opened 2022-07-01
Revert "fix: duplicate metadata output from chores"

victorlin:
Reverts nextstrain/monkeypox#87 now that https://github.com/nextstrain/augur/issues/985 is fixed as of Augur version 16.0.2.


victorlin:
@huddlej

PR #93, opened 2022-07-07
Subsample MPXV clades

trvrb:

As more and more B.1 viruses are sequenced, the MPXV global diversity tree is becoming increasingly lopsided and not very descriptive of broader diversity.

This commit updates filter to subsample based on `clade lineage` so that every clade / lineage combination is subsampled to. It sets sequences per group to 1000 for hMPXV so that no subsampling occurs and sets sequences per group to 100 for MPXV, which effectively just subsamples B.1 from 289 genomes to 100.

Here's what the MPXV build looks like with this subsampling in place:

Screen Shot 2022-07-06 at 5 44 58 PM

Build visible here: https://next.nextstrain.org/staging/monkeypox/mpxv/subsample-mpxv-clades

I think this subsampling will be increasingly important for the `mpxv` page as more B.1* genomes come in.

### DIFF
--- a/config/config_hmpxv1.yaml
+++ b/config/config_hmpxv1.yaml
@@ -17,6 +17,7 @@ auspice_name: "monkeypox_hmpxv1"
 ## filter
 min_date: 2017
 min_length: 10000
+sequences_per_group: 1000
 filters: "--exclude-where clade!=hMPXV-1"
 
 ## align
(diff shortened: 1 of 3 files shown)


rneher (approved):
Thanks. makes sense!

trvrb:
Thanks @rneher! I'll merge and rebuild.

PR #95, opened 2022-07-23
clades: Proposal to designate 5 B.1 sublineages (B.1.1 - B.1.5)

corneliusroemer:
## Proposal to designate 5 B.1 sublineages

Lineages are defined by at least one shared nucleotide mutation on top of the B.1 polytomy.

In this first round of sublineage designations, the clusters with the highest number of sequences publicly available via Genbank are chosen.

It is important to bear in mind that the proportion of cases sequenced with sequences shared via Genbank differs hugely among countries. Germany has sequenced ~10% of their known cases. While for instance the Netherlands have only shared one sequence publicly, for 500 known cases (0.2%), a proportion 50x smaller than Germany's.

As a result, lineages are disproportionately associated with those countries that sequence most (currently Germany and Canada). Lineages should not be used on their own without other metadata to draw epidemiological conclusions.
 
These lineages have been designated to enable researchers to track the virus on a finer scale. Their designation does not imply any biological difference. For example, B.1.4 differs from B.1 only by a synonymous mutation - the proteins are thus all identical.

### Testing

Staging build deployed here: https://nextstrain.org/staging/monkeypox/hmpxv1/lineage-proposal?label=clade:B.1&m=div

### DIFF
--- a/config/clades.tsv
+++ b/config/clades.tsv
@@ -43,9 +43,3 @@ B.1.4	nuc	34308	A
 
 B.1.5	clade	B.1
 B.1.5	nuc	70780	T
-
-B.1.6	clade	B.1
-B.1.6	nuc	25644	T
-
-B.1.7	clade	B.1
-B.1.7	nuc	14611	A
--- a/config/clades.tsv
+++ b/config/clades.tsv
@@ -36,16 +36,16 @@ B.1.2	clade	B.1
 B.1.2	nuc	186165	A
 
 B.1.3	clade	B.1
-B.1.3	nuc	34308	A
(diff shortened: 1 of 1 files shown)


emmahodcroft:
I think this is a great start, would be great to hash out the criteria more directly and decide where to draw lines. Most of my thoughts/discussion is on Slack!

corneliusroemer:
Merge after discussion and agreement within committee

AngieHinrichs:
Just checking - will these be added to nextclade too?

corneliusroemer:
Yes, they are on master.clades.nextstrain.org already - about to release ;)

Issue #36, opened 2022-06-07
ingest: thread Slack notifications


joverlee521:
### Context
We get multiple Slack notifications per ingest run, so it would be cleaner to have these notifications threaded. 

### Possible Solution
- See [ncov workflow](https://github.com/nextstrain/ncov/blob/b94913b4791d0095bfa857f7b2a7fb2081480782/workflow/snakemake_rules/export_for_nextstrain.smk#L384-L415) for example of how this can be done with `PersistentDict`.

- Maybe another way we can go about this is edit the global Snakemake `config` to store the `thread_ts`? (Just a thought, haven't really tested if this is possible...)

jameshadfield:
> Maybe another way we can go about this is edit the global Snakemake config to store the thread_ts

This dosn't work (unfortunately) because each rule gets a copy of the config, which prevents this message-passing approach. `PersistentDict` was the only solution I could get to work (short of writing to a small file), but it's tricky to debug as when running locally a new snakemake run will still have access to the previous dict which can get confusing (it works for automated runs as they're in a new worker / environment).

PR #100, opened 2022-08-02
Automate rebuilds

joverlee521:
Automates rebuilds for `mpxv` and `hpmxv1` to run when there's new data from ingest. Includes threaded Slack replies for builds since we don't want to mix up Slack messages from the separate builds. (Note this is separate from #36, which is specifically for ingest). 

### Testing
Unfortunately cannot fully test the ingest trigger since [repository dispatch events](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch) only trigger the workflow on the main branch.

### DIFF
--- a/.github/workflows/rebuild-hmpxv1.yaml
+++ b/.github/workflows/rebuild-hmpxv1.yaml
@@ -1,0 +1,47 @@
+name: Rebuild hmpxv1
+
+on:
+  repository_dispatch:
+    types:
+      - rebuild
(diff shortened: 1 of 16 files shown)


corneliusroemer:
@joverlee521 is it intended that the output of the auto-builds only contains around 20 sequences? I downloaded the output tree and put it in Auspice.

Looks good so I'll merge so we can see how it goes - not much can go wrong here I think - in worst case we'll revert ;)

PR #101, opened 2022-08-08
Add clade reference sequences

corneliusroemer:
This PR adds a file that lists one or multiple (in case there is no basal one) reference sequences for each clade/lineage combination.

This should be useful to make minimal trees that display all known clades/lineages.

The sequences were chosen to be:
1. Available via NCBI/Genbank
2. As complete as possible (few terminal and internall Ns or ambiguous nucleotides)
3. As basal as possible (or if no basal one is available, multiple sequences are chosen so that the MRCA is basal)

There is often no single best choice - better sequences may become available over time in which case the representative sequences could change.

This is a first draft, so the format may still change.

I did not want to make strain/isolate name something that we rely on since it often changes - hence I have packed it in the "comment" column. It may also not always exist independently of accession.

We could potentially add a convenience link to the Genbank page for each sequence - but this is somewhat redundant since we have the accession.

Original proposal to implement this came from @rambaut


emmahodcroft:
Shall we move this PR to https://github.com/mpxv-lineages/lineage-designation now? I think the file may be more appropriate there, esp as it's not used in the build itself.

corneliusroemer:
@emmahodcroft yes good idea - though I'm a bit worried that people may start relying on this.

Maybe we can put "experimental" files in a special folder that says "experimental"? Or add it to the file name?

emmahodcroft:
We can just let it be a PR over there, and not merge it yet? I suppose the clades (2, 3) need updating (in line with whenever we update everything else). I guess more generally, there's still some uncertainty about ensuring the right references have been chosen?

corneliusroemer:
Ported to https://github.com/mpxv-lineages/lineage-designation

PR #105, opened 2022-08-19
feat(ingest): add outbreak column from Nextclade output

corneliusroemer:


corneliusroemer:
Tested and it seems to work!

AngieHinrichs:
How do you get nextclade to produce the `outbreak` column?  I have nextclade 2.4.0 and don't see it:
```
nextclade dataset get --name hMPXV --output-zip hMPXV.zip
time nextclade run \
    -D hMPXV.zip \
    -j 30 \
    --output-tsv nextclade.tsv \
    --output-fasta nextalign.fa.xz \
    --retry-reverse-complement true \
    genbank.fa.xz
```

corneliusroemer:
Strange @AngieHinrichs - could you share the exact files? The dataset zip and the Genbank fa xz? I'll try to reproduce.

My guess is you're not using the newest dataset? Maybe some caching issue? That would be the only sensible explanation I can think of.

AngieHinrichs:
My mistake, I was using a hMPXV.zip downloaded about 10 hours ago.  Re-downloading it did the trick, sorry about the false alarm.

corneliusroemer:
So you didn't actually run the code you shared to reproduce 🤣  Classic problem - you think you know what you did when in fact there's a subtle difference :)

PR #106, opened 2022-08-20
make groupby build specific

rneher:
we need the grouping criteria to be specific to mpxv and hmpxv1.


config/config_hmpxv1.yaml
```diff
@@ -17,7 +17,8 @@ auspice_name: "monkeypox_hmpxv1"
 ## filter
 min_date: 2017
 min_length: 10000
-sequences_per_group: 1000
+sequences_per_group: 20
+groupby: "clade lineage year month country"
```
corneliusroemer, line 21: Not sure I'm in favour of this as it means discovering new lineages will be harder.

This will downsample mostly B.1's that are not assigned any lineage.

What if we leave out the lineage here? Clade doesn't do anything either so can be deleted as well, leaving us with `year month country`, we could then adjust this to say 50 to reduce the amount of sequences from Germany.

rneher: Yes, we could reduce this to `year month country`

rneher:
this adjusts the filtering to have around 700 strains in the hmpxv1 build (downsizing germany slightly).

The mpxv filtering now always includes strains from all countries. In the previous one countries with few genomes often went missing.

workflow/snakemake_rules/core.smk
```diff
@@ -44,7 +44,7 @@ rule filter:
         metadata = build_dir + "/{build_name}/metadata.tsv",
         log = build_dir + "/{build_name}/filtered.log"
     params:
-        group_by = "clade lineage",
+        group_by = config.get("groupby", "clade lineage"),
         sequences_per_group = config['sequences_per_group'],
         min_date = config['min_date'],
         min_length = config['min_length'],
```
victorlin, line 50: Being nit-picky here. Would be nice to make config key == param key, to be (mostly) consistent with what's below.

```suggestion
        group_by = config.get("group_by", "clade lineage"),
        sequences_per_group = config['sequences_per_group'],
        min_date = config['min_date'],
        min_length = config['min_length'],
```

rneher: makes sense. Thanks.

PR #110, opened 2022-08-24
feat: mask more specifically

corneliusroemer:

Sometimes it seems that IQtree splits branches into two based on clusters of Ns/not-Ns.

I carefully reviewed alignments and masked all tandem repeat and homopolymers stretches where a significant portion of sequences have Ns.

This should make trees more stable.

I tested builds for both hmpxv1 and mpxv with the new mask.

I think some of the previous masking was done against a different reference, hence some of our mask regions may have been shifted slightly. This is fixed now.

### DIFF
--- a/config/config_hmpxv1.yaml
+++ b/config/config_hmpxv1.yaml
@@ -32,6 +32,6 @@ root: "MK783032 MK783030"
 recency: true
 
 mask:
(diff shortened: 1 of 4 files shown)


corneliusroemer:
This should help avoid these kinds of issues in our current build:
image

PR #113, opened 2022-08-28
add steps to merge sister branches with shared mutations

rneher:
we repeated run into situations like this:
image

This is a problem of IQtree when it hits large polytomies. this PR introduces a step where sister branches with shared mutations are merged.

### DIFF
--- a/scripts/fix_tree.py
+++ b/scripts/fix_tree.py
@@ -1,0 +1,42 @@
+from collections import defaultdict
+import argparse
+from treetime import TreeAnc
+from Bio import Phylo
+
+if __name__=="__main__":
+    parser = argparse.ArgumentParser(
+        description="remove time info",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
+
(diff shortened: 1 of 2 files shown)


corneliusroemer:
Great to have a script that does this! Could be useful for all builds if this works robustly.

Should we merge this and just be careful to check trees for the next few days to make sure there's no bug?

rneher:
would be good to test a bit more on a case where this is an issue. So far, I just hacked this together (also fairly inefficient). But otherwise then I think soon good to merge.

corneliusroemer:
We could make dual builds - one with one without so we can compare.

corneliusroemer:
Test data for a broken build that gets improved (`C25644T` no longer obviously wrong) but not totally fixed (`G42504A` homomplasy inside the merged parts - so probably we need to run the program on subtrees of merged parents to fix merged ones up.

image

[test_data_tree_fix.tar.zst.txt](https://github.com/nextstrain/monkeypox/files/9491266/test_data_tree_fix.tar.zst.txt)

PR #114, opened 2022-08-28
add extra nextclade rule to align against the proper reference, uploa…

rneher:
Some people said it would be helpful to have the aligned sequences, translations, and insertions available for download. This PR adds this.

### DIFF
--- a/ingest/workflow/snakemake_rules/nextclade.smk
+++ b/ingest/workflow/snakemake_rules/nextclade.smk
@@ -20,7 +20,7 @@ rule align:
         sequences = "data/sequences.fasta",
         dataset = "hmpxv.zip"
     output:
-        alignment = "data/alignment.fasta.xz",
+        alignment = "data/alignment.fasta",
         insertions = "data/insertions.csv",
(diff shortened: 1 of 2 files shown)


ingest/workflow/snakemake_rules/nextclade.smk
```diff
@@ -7,6 +7,33 @@ rule nextclade_dataset:
...
+        temp("hmpxv.zip")
+    shell:
+        """
+        nextclade dataset get --name hMPXV --output-zip {output}
+        """
+
+rule align:
+    input:
+        sequences = "data/sequences.fasta",
+        dataset = "mpxv.zip"
```
joverlee521, line 21: Is this supposed to be `hmpxv.zip`? Only asking because the hMPXV dataset was added above but doesn't look like it's used anywhere.

rneher: thanks, Jover. align now uses hmpxv. shouldn't matter much in practice, but I think people will be more confidence when the alignment is based on an actual sequence, rather than a reconstructed ancestor.

PR #119, opened 2022-09-06
improve script to fix trees

rneher:
this makes the fix tree script iterative and retains the proper substructure of newly created clades through merging,


rneher:
@corneliusroemer this looks like it is working. Please have a look and merge

corneliusroemer:
Just running it on the suboptimal tree from yesterday - will then merge. It'll remain for testing on the big build - but we could enable fixing any time on the other builds.

corneliusroemer:
Excellent logging - works very well on the build - yesterday it only did iteration one, now it goes deeper.

image

corneliusroemer (approved):
Tested and works well

PR #121, opened 2022-09-06
feat: include all sequences from non-B.1 in hmpxv1 build

### LINKED ISSUE #120: ENH: Don't subsample non-B.1 lineages
### Context
Right now we may sample out some non-B.1 sequences because I argued we shouldn't sample on `lineage country month year` only on `country month year` -> my mistake.

We should sample by country within B.1 - but not subsample outside B.1. That way we combine the best of both.

Raised by @rambaut

corneliusroemer:
Resolves #120 

This PR adds all non-B.1-hmpxv1 sequences in the hmpxv1 build. 

The implementation is possibly not ideal but I iteratively ended up there.

I first create a list of strains to include using `tsv-filter` based on metadata.

I then exclude all sequences in exclude using augur-filter. I then exclude-all and include the list generated in the first step.

I then concatenate the ordinary filter output and deduplicate metadata and sequences using tsv-uniq and seqkit rmdup, respectively.

Testing: I ran this locally and it worked.


rneher:
I understand that just grouping by lineage has its downsides (B.1 shrinks disproportionally), but this adds a lot of bespoke complexity.

corneliusroemer:
Good point @rneher. We may just add all non-B.1s to all builds - the effect would be identical with less complexity.

Would you call deduplication complexity? I'll merge metadata and sequence deduplication rules to reduce overall rule numbers.

corneliusroemer:
This should be much better now @rneher - we simply include all good As in all builds. This is what we've been doing in MPXV and hmpxv1_big anyways (due to lineage sampling and no-sampling respectively).

corneliusroemer:
That would be find for now but as more sequences arrive we start subsampling only B.1 and not the (somewhat arbitrarily defined) B.1.X which would not really be nice in terms of making the build representative.

I guess there are four goals here:
a) country (not overly distort towards big submitters)
b) lineage (not distort towards designated lineages)
c) not sample undersampled, that is non B.1*
d) keep code minimal

My proposal is bad re minimal code, that is a)

Right now we don't satisfy c)

Your proposal would go against b)

rneher:
looking at your code a little closer, wouldn't it be easier to generate an include list that contains all `A*` sequences that are not in `exclude.txt`

as in 
```
with open('include.txt') as fh:
    include_list = set([x.strip() for x in fh])
with open('exclude.txt') as fh:
    exclude_list = ([x.strip().split()[0] for x in fh])

with open('pruned_include.txt`, 'w') as fh:
    fh.write('\n'.join(include_list.difference(exlude_list) + '\n')
```
I guess your version works. But this should work too, right? if so, seems easier than the backward/forward filtering, tmp files, deduplications, etc

corneliusroemer:
I thought about that option @rneher but it's a static include if I'm not mistaken, which means we need to update it whenever there are new As, which is not very nice.

We could reduce number of extra rules by adding the deduplication to the filter rule.

Possibly cleaner would be to create the include list first (dynamically) and then pass it to the second filter rule. That way no deduplication/joining is necessary.

This may be the best I think. I'll quickly try it out.

corneliusroemer:
I think I've now converged on something that should be acceptable - can you have a look @rneher?

rneher:
this is cleaner. But I think you misunderstood me. my idea was:

 * make an include list with all As, as you do now
 * reduce that include list to exclude bad strains
 * use that list of good As as include in filter.

corneliusroemer:
Ah yes I see - so you meant the thing that I ended up implementing.

I think what you propose wouldn't work since we don't just need to exclude the exclude.txt strains but also too short and wrongly dated ones.

As far as I'm aware there's no "include despite subsampling, but let other excludes still apply, like `--min-date`, `--min-length`"

corneliusroemer:
An alternative would be to filter strains using:
```
--query "lineage < 'B' & lineage >='A'"
```

This could combine: `exclude_bad` and `create_A_strain_list` in one rule, but then one still would need to combine the resulting two sets of sequences.

So I guess I'll stick with what we have for now.

workflow/snakemake_rules/core.smk
```diff
@@ -28,11 +28,10 @@ rule wrangle_metadata:
         """
 
 
-rule filter:
+rule exclude_bad:
```
joverlee521, line 31: I think it would be clearer to call this `rule filter`

workflow/snakemake_rules/core.smk
```diff
@@ -41,28 +40,70 @@ rule filter:
...
+        """
+        tsv-filter -H \
+            --str-in-fld lineage:'A'\
+            {input.metadata} \
+        | tsv-select -f 1 \
+        > {output.include_strains}
+        """
+
+
+rule filter:
```
joverlee521, line 79: I think it would be clearer to call this `rule subsample`

workflow/snakemake_rules/core.smk
```diff
@@ -41,28 +40,70 @@ rule filter:
...
+    message:
+        """
+        Filtering to
+          - {params.sequences_per_group} sequence(s) per {params.group_by!s}
+        """
+    input:
+        sequences=rules.exclude_bad.output.sequences,
+        metadata=rules.exclude_bad.output.metadata,
+        include=rules.include_A_strains.output.include_strains,
```
joverlee521, line 88: It would be easier for people to add their own custom rules if we keep all of inputs as the file path rather than the rule outputs.

workflow/snakemake_rules/core.smk
```diff
@@ -41,28 +40,70 @@ rule filter:
...
         sequences=build_dir + "/{build_name}/filtered.fasta",
         metadata=build_dir + "/{build_name}/metadata.tsv",
-        log=build_dir + "/{build_name}/filtered.log",
+        log=build_dir + "/{build_name}/filter.log",
     params:
         group_by=config.get("group_by", "clade lineage"),
         sequences_per_group=config["sequences_per_group"],
-        min_date=config["min_date"],
-        min_length=config["min_length"],
         other_filters=config.get("filters", ""),
```
joverlee521, line 96: Should this get moved up to the other filter rule?

joverlee521 (commented):
The two step filter + subsample makes sense to me!

Even though the PR has been merged, I left a couple comments to make the workflow easier to use for the outside community.

PR #123, opened 2022-09-07
feat: add tree_exclude sites so that Colombian sequences don't need to be excluded

corneliusroemer:
There are 2 Colombian sequences with 4 reversions to A.1.

These mess up tree topology. So we've excluded them for now.

A better way is to mask these 4 sites (they are in ITR so not surprising there can be issues there) for tree build only (soft mask). That way the tree topology is correct but the sequences show still on the tree.

Test run here: should appear on staging in 30-45 min  for checking
https://github.com/nextstrain/monkeypox/actions/runs/3007463882

### DIFF
--- a/config/config_hmpxv1.yaml
+++ b/config/config_hmpxv1.yaml
@@ -7,6 +7,7 @@ clades: "config/clades.tsv"
 lat_longs: "config/lat_longs.tsv"
 auspice_config: "config/auspice_config_hmpxv1.json"
 description: "config/description.md"
+tree_mask: "config/tree_mask.tsv"
 
 strain_id_field: "accession"
 display_strain_field: "strain_original"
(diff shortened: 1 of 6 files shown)


rneher (approved):

AngieHinrichs:
I'm curious, why add a new config/tree_mask.tsv instead of adding the sites to mask.bed and mask_overview.bed?

AngieHinrichs:
Never mind, now I see this part of the description:

> That way the tree topology is correct but the sequences show still on the tree.

corneliusroemer:
Yes indeed - if we masked in mask.bed, it would end up being totally masked hence shortening the branches and reducing the clock rate.

emmahodcroft:
Yes sorry for not answering @AngieHinrichs! Using a separate file, we can exclude sequences only from the tree topology inference, but still use them for other steps (branch length, ancestral reconstruction). This is a nice solution for sites that are right in most sequences (they will still be informative for branch length in most of the tree & will be reconstructed appropriately most places) but may be messing up the topology somewhere (like false reversions). If it site itself is a problem (in a repeat, repeated primer error, area of generally bad coverage like start & end) then instead, we'd just exclude it from ever being used for anything.

PR #131, opened 2022-10-03
Cap the tree rule's threads to whatever was given to the workflow

tsibley:
This solves two issues:

 1. It informs Snakemake's scheduler about the number of threads this rule will use.  Previously, the rule would use as many as possible but Snakemake's scheduler would think it was only using 1, potentially leading the scheduler could oversubscribe the system. This was likely only a latent issue given the current serial nature of the workflow.

 2. It prevents Augur's CPU detection from seeing all the host CPUs even when it's running inside a CPU-time limited container context.  See also <https://github.com/nextstrain/augur/issues/1058>.

When I briefly audited the rest of the workflow for implicit threads usage, I noted that this same pattern is already used by the align rule.

### DIFF
--- a/workflow/snakemake_rules/core.smk
+++ b/workflow/snakemake_rules/core.smk
@@ -189,6 +189,7 @@ rule tree:
         tree_mask=config["tree_mask"],
     output:
         tree=build_dir + "/{build_name}/tree_raw.nwk",
+    threads: workflow.cores
     shell:
         """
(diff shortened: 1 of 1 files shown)


tsibley:
Going to merge pre-review because I believe it's correct, the same pattern is already used for `nextalign`, and we'll find out easily enough if it's not.

PR #140, opened 2022-12-09
Update PAT_GITHUB_DISPATCH secret

joverlee521:
Use the shared organization secret for the dispatch token so it no longer has to be managed as a separate repo secret.

I've already added the org secret `GH_TOKEN_NEXTSTRAIN_BOT_WORKFLOW_DISPATCH` to this repo.

### DIFF
--- a/.github/workflows/rebuild-all.yaml
+++ b/.github/workflows/rebuild-all.yaml
@@ -6,13 +6,12 @@ on:
 jobs:
   rebuild_all:
     runs-on: ubuntu-latest
-    env:
-      GITHUB_RUN_ID: ${{ github.run_id }}
-      SLACK_CHANNELS: monkeypox-updates
-      SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
     steps:
+      - uses: actions/checkout@v3
       - name: Repository Dispatch
-        uses: peter-evans/repository-dispatch@v2
-        with:
-          token: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_WORKFLOW_DISPATCH }}
-          event-type: rebuild
+        run: ./ingest/bin/trigger monkeypox rebuild
+        env:
+          GITHUB_RUN_ID: ${{ github.run_id }}
+          SLACK_CHANNELS: monkeypox-updates
+          SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
+          PAT_GITHUB_DISPATCH: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_WORKFLOW_DISPATCH }}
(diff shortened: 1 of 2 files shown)


.github/workflows/rebuild-all.yaml
```diff
@@ -14,5 +14,5 @@ jobs:
       - name: Repository Dispatch
         uses: peter-evans/repository-dispatch@v2
         with:
-          token: ${{ secrets.PAT_GITHUB_DISPATCH }}
+          token: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_WORKFLOW_DISPATCH }}
```
joverlee521, line 17: I didn't realize we were using a third party action to create a repo dispatch until now. Should we be worried about passing org secrets to a third party action here?

victorlin, line 17: From repository dispatch docs:

> This endpoint requires write access to the repository by providing either:
> - Personal access tokens with `repo` scope. For more information, see "[Creating a personal access token for the command line](https://docs.github.com/articles/creating-a-personal-access-token-for-the-command-line)" in the GitHub Help documentation.

Although the workflow dispatch token has implicit `repo` scope, it might be better to instead use the `GH_TOKEN_NEXTSTRAIN_BOT_REPO` token which has `repo` scope alone.

joverlee521: Updated in f33b6d1836471dcd531d735ba0d9e8c7566d3d0a to stop using a third-party action.

joverlee521: Hmm, now I'm confused why `GH_TOKEN_NEXTSTRAIN_BOT_WORKFLOW_DISPATCH` has additional scope if it's only used to dispatch other GitHub actions workflows.

From [workflows dispatch event docs](https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event):
> You must authenticate using an access token with the `repo` scope to use this endpoint.

victorlin: Oh, good point... I don't have a good reason. We can remove the `workflow` scope and see if it still works.

joverlee521: I removed the `workflow` scope and tested with https://github.com/nextstrain/test-github-actions/actions/runs/3660631844. It was able to trigger a [docker-base CI run ](https://github.com/nextstrain/docker-base/actions/runs/3660632757) which I cancelled.

joverlee521: Merging this now to fix our automated runs. We can revisit later on whether we should consolidate the two tokens that now have the same `repo` scope.

joverlee521:
Deleted the repo secret PAT_GITHUB_DISPATCH.

PR #143, opened 2023-02-03
fix: Use first field instead of last field

j23414:

Assuming the documentation in transform strain names is the goal

> If multiple fields are provided, will use the first field that has a non-empty string.

Need a `break` in the for loop, otherwise "last field" will be used.

Ran across this because I was exploring a ncbi datasets genbank.ndjson file:

```
cat genbank.ndjson  \
  | ./bin/transform-strain-names  \
    --strain-regex ^.+$  \
    --backup-fields "Isolate Lineage" "Accession" "accession" \
  > genbank_transform_strain_name.ndson
```

### Testing

- [x] Checks pass

### DIFF
--- a/ingest/bin/transform-strain-names
+++ b/ingest/bin/transform-strain-names
@@ -40,6 +40,7 @@ if __name__ == '__main__':
                 for field in args.backup_fields:
                     if record.get(field):
                         record['strain'] = str(record[field])
+                        break
 
         if record['strain'] == '':
             print(f"WARNING: Record number {index} has an empty string as the strain name.", file=stderr)


j23414:
Might only be relevant if multiple fields are provided here

https://github.com/nextstrain/monkeypox/blob/0ffc8d8451a4250aea95a9240551401575c0f6e6/ingest/config/config.yaml#L14

joverlee521 (approved):
Thanks for catching this @j23414! This is indeed needed for the script to work as described. 

The [CI/lint job](https://github.com/nextstrain/monkeypox/actions/runs/4088842347/jobs/7050917896) is unhappy because of some formatting issues unrelated to this change. Looks like your branch is [4 commits behind master](https://github.com/nextstrain/monkeypox/compare/0ffc8d8451a4250aea95a9240551401575c0f6e6...b32faf325479e354f9b04ff0b2aee93ac4c65d04) and these commits have formatting updates that would resolve this error. You can resolve the CI error for this PR by rebasing your branch onto the current master branch.

PR #145, opened 2023-02-06
ingest: Add `QC_overall_status` column

joverlee521:
Allows downstream users of the public metadata TSV to filter sequences based on the overall QC status output from Nextclade.

### DIFF
--- a/ingest/bin/join-metadata-and-clades.py
+++ b/ingest/bin/join-metadata-and-clades.py
@@ -20,6 +20,7 @@ column_map = {
     "qc.privateMutations.status": "QC_rare_mutations",
     "qc.frameShifts.status": "QC_frame_shifts",
     "qc.stopCodons.status": "QC_stop_codons",
+    "qc.overallStatus": "QC_overall_status",
     "frameShifts": "frame_shifts",
     "isReverseComplement": "is_reverse_complement",
 #    "deletions": "deletions",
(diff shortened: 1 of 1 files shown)


joverlee521:
@corneliusroemer is there a reason why this column isn't already included in the output? 

This was requested internally by a Bedford lab grad student who only wants to work with good quality sequences. I just wanted to double check with you that this wasn't excluded for some particular reason.

corneliusroemer:
I faintly remember that we may have done this because the `QC_overall_status` is a heuristic that is brittle and may depend on which sequences you're looking at in particular. There is probably no one-size-fits-all for good vs bad.

When you look at sequences from less sampled clades what is bad may require a very different weighting of the QC constituents.

It may be best for the grad student to make a conscious decision of what is good/bad by looking at the sequences and individual QC metrics.  Does that make sense?

joverlee521:
> I faintly remember that we may have done this because the QC_overall_status is a heuristic that is brittle and may depend on which sequences you're looking at in particular. There is probably no one-size-fits-all for good vs bad.
>
> When you look at sequences from less sampled clades what is bad may require a very different weighting of the QC constituents.

Ah, I see. We use this column in SARS-CoV-2 so I didn't realize this was potentially brittle. I assume this is because there are much less monkeypox sequences than SARS-CoV-2 sequences so it's less reliable for monkeypox?

corneliusroemer:
> Ah, I see. We use this column in SARS-CoV-2 so I didn't realize this was potentially brittle. I assume this is because there are much less monkeypox sequences than SARS-CoV-2 sequences so it's less reliable for monkeypox?

It's not necessarily brittle, it'll output something. But given the different time scales of evolution and resulting divergences between a few nuc mutations and 1000 it's hard to be meaningful across the scale.

The overall QC makes more sense in SC2 where it's fairly well-tuned. Even then, it shouldn't be overinterpreted there.

PR #146, opened 2023-02-08
Parameterize ncbi_id in fetch_sequences

j23414:

Could we parameterize the NCBI Taxon ID in the fetch sequences scripts? 

Historically the NCBI Taxon ID has been hardcoded in `ingest/bin/genbank-url`:

https://github.com/nextstrain/monkeypox/blob/b54768ec17872eb0d898e29527785642f6b98c0d/ingest/bin/genbank-url#L21

This PR generalizes that and related scripts such that the NCBI Taxon ID can be defined in a snakemake rule:

https://github.com/nextstrain/monkeypox/blob/b54768ec17872eb0d898e29527785642f6b98c0d/ingest/workflow/snakemake_rules/fetch_sequences.smk#L17-L23

[View example here! 👀 ](https://github.com/j23414/monkeypox/blob/5fd5910f3842373afeaf7dc288845c4b3545583b/ingest/workflow/snakemake_rules/fetch_sequences.smk#L17-L23)

Open to discussion or other ways to parameterize this. Can test the code via:

```
git clone https://github.com/j23414/monkeypox.git
cd monkeypox/ingest
nextstrain build  . data/sequences.ndjson
```

or simpler

```
./bin/fetch-from-genbank 10244 > genbank.ndjson
```

### Related issue(s)

Related commit: https://github.com/nextstrain/dengue/pull/6/commits/5d3281dd1a4f434a46bf16e2ea951f510c0cec8e

### Testing

- [ ] Checks pass

### DIFF
--- a/ingest/bin/genbank-url
+++ b/ingest/bin/genbank-url
@@ -12,54 +12,80 @@ and observing the network activity at
     https://www.ncbi.nlm.nih.gov/labs/virus/vssi/#/virus?SeqType_s=Nucleotide&VirusLineage_ss=Monkeypox%20virus,%20taxid:10244
 """
 from urllib.parse import urlencode
+import argparse
 
-endpoint = "https://www.ncbi.nlm.nih.gov/genomes/VirusVariation/vvsearch2/"
-params = {
-    # Search criteria
-    'fq': [
-        '{!tag=SeqType_s}SeqType_s:("Nucleotide")', # Nucleotide sequences (as opposed to protein)
-        'VirusLineageId_ss:(10244)',                # NCBI Taxon id for Monkeypox
-    ],
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Given an NCBI taxon ID, generate URL to download "
+        "all viral sequences and their curated metadata from GenBank via NCBI Virus."
+    )
+    parser.add_argument(
+        "--ncbi-taxon-id",
+        help="NCBI Taxon ID.",
+        default="10244",
+        required=True
+    )
+    return parser.parse_args()
 
-    # Unclear, but seems necessary.
-    'q': '*:*',
+def build_query_url(ncbi_taxon_id: str):
+    """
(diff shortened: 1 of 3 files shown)


j23414:
Speaking of `NCBI datasets`, parameterizing the NCBI Taxon ID feeds in nicely to the `datasets` command:

```
 f"datasets download virus genome taxon {params.ncbi_id} --no-progressbar --filename {output.dataset_package}
```

https://github.com/nextstrain/ncov-ingest/blob/master/workflow/snakemake_rules/fetch_sequences.smk#L53

ingest/bin/genbank-url
```diff
@@ -12,54 +12,85 @@ and observing the network activity at
...
+    """
+    parser = argparse.ArgumentParser(
+        description="Given an NCBI taxon ID, generate URL to download "
+        "all viral sequences and their curated metadata from GenBank via NCBI Virus."
+    )
+    parser.add_argument(
+        "--taxonid",
+        help="NCBI Taxon ID.",
+        required=True
+    )
```
tsibley, line 29: _non-blocking_

`--taxon-id` would be a more canonical option name.

Maybe default to 10244 so that any existing callers outside of this repo don't have to change?

tsibley (changes requested):
+1 for the parameterization. I've left a number of specific comments on small things which I think are important enough to mention, but none of them I consider blocking (e.g. if for some reason you can't/won't address them).

Additionally, I think it would be good practice to put some of the rationale you included in your PR description into the commit message itself, as that's more durable and tightly linked to the source code.

ingest/bin/genbank-url
j23414: Agree! Basically tried to convert all parameter names to "ncbi-taxon-id" and set default to 10244. 

Thanks for helpful suggestions and let me know if you catch anything else!

tsibley (approved):

PR #147, opened 2023-03-28
fix: fix and rewrite the help description

j23414:

While responding to

* https://github.com/nextstrain/dengue/pull/6#discussion_r1149854478

Fixes the incorrect docstring.

### Testing

Since this is documentation, does not affect data processing.

- [x] Checks pass


tsibley (approved):
Thanks!

Note that "docstring" has [a specific meaning](https://peps.python.org/pep-0257/#what-is-a-docstring). This here is a string used in documentation (specifically the `--help` output of this program), but it's not a docstring.

PR #149, opened 2023-04-12
fix: Add `--cores all` to rebuild-mpxv.yaml

corneliusroemer:
This is required for the action to be compatible with snakemake version >5.11.0
which started to require the `--cores` option

For more context see:
- https://github.com/nextstrain/cli/issues/272 
- https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1681324279363569?thread_ts=1681293140.124269&cid=C01LCTT7JNN

- [ ] Checks pass

I added `--cores all` because we don't limit number of cores to use in the call of `nextstrain build`

### DIFF
--- a/.github/workflows/rebuild-mpxv.yaml
+++ b/.github/workflows/rebuild-mpxv.yaml
@@ -35,6 +35,7 @@ jobs:
           . \
             envdir env.d snakemake notify_on_deploy \
               --configfiles config/config_mpxv.yaml config/nextstrain_automation.yaml \
+              --cores all \
               --printshellcmds
       env:
         AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}


.github/workflows/rebuild-mpxv.yaml
```diff
@@ -35,6 +35,7 @@ jobs:
           . \
             envdir env.d snakemake notify_on_deploy \
               --configfiles config/config_mpxv.yaml config/nextstrain_automation.yaml \
+              --cores all \
```
tsibley, line 38: Hmm, I don't think we should use `--cores all` because of https://github.com/nextstrain/cli/issues/175. (I mentioned this earlier, but I might not have explained it well then.)

corneliusroemer: I didn't use `all` whenever there was an explicit number given. I didn't know what to pass otherwise when `--cpus` wasn't passed.

What does `nextstrain build` use as `--cpus` by default? CLI `--help-all` said `None`? So if there's no machine up it won't run? Or will it be like 0.00001?

What would you put instead in case of no `--cpus`?

.github/workflows/rebuild-hmpxv1-big.yaml
```diff
@@ -37,6 +37,7 @@ jobs:
             . \
               envdir env.d snakemake notify_on_deploy \
                 --configfiles config/config_hmpxv1_big.yaml config/nextstrain_automation.yaml \
+                --cores 8
```
corneliusroemer, line 40:
```suggestion
                --cores 8 \
```

.github/workflows/fetch-and-ingest.yaml
```diff
@@ -55,6 +55,7 @@ jobs:
           ingest \
             envdir env.d snakemake \
               --configfiles config/config.yaml config/optional.yaml \
+              --cores all
```
corneliusroemer, line 58:
```suggestion
              --cores all \
```

.github/workflows/rebuild-mpxv.yaml
tsibley: It uses the default as configured in AWS Batch. That will vary for different organizations, but for us, it's currently 4.

For this case above, I would explicitly pass `--cpus 4` to `nextstrain build` to make the AWS Batch default explicit (until https://github.com/nextstrain/cli/issues/175 is fixed) and then also `--cores 4` to `snakemake` (until https://github.com/nextstrain/cli/issues/272 is fixed, probably via [first-class env var management support](https://github.com/tsibley/blab-standup/blob/master/2022-08-26.md)).

tsibley: I mean, ok, I guess, but I don't see how `--cores all` here is in any way better or quicker than `--cpus 4` + `--cores 4`.

tsibley: To make what I'd said earlier very clear: if we run concurrent workflows with small numbers of requested CPUs, they are likely to get packed onto the same machine, where overscheduling jobs by Snakemake is likely to make them all slower. Slow enough to notice? Maybe, maybe not. I don't know, as I rarely pay attention to our workflow execution times.

PR #150, opened 2023-04-13
Snakefile: Set AUGUR_RECURSION_LIMIT=10000

joverlee521:

This change is a response to the recursion error that a user ran into when running the workflow with a large number of sequences during Nextstrain office hours on 2023-04-13.

Set the recursion limit globally instead of asking the user to remember to set the environment variable or edit their own Snakefile.

### Testing

- [x] Checks pass

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -8,6 +8,13 @@ if version.parse(augur_version) < version.parse(min_version):
     print(f"Current augur version: {augur_version}. Minimum required: {min_version}")
     sys.exit(1)
 
+# Set the maximum recursion limit globally for all shell commands, to avoid
+# issues with large trees crashing the workflow.  Preserve Snakemake's default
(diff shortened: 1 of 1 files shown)


joverlee521:
We've never seen this issue with the Nextstrain automated builds in AWS Batch because the [`nextstrain-job` definition](https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#job-definition/ec2/detail/arn:aws:batch:us-east-1:827581582529:job-definition/nextstrain-job:25) already sets the `AUGUR_RECURSION_LIMIT` to 10000.

tsibley (approved):
LGTM.

Separately, we might also revisit adding to Augur a default bump of the recursion limit. I think the question in the past has been, "what would that be?". The answer might be 10,000 if that's a value we're broadly successful with (at least for now).

PR #151, opened 2023-04-14
Connect the output sequence parameter

j23414:

This bug was invisible before because when the flag `--sequence` is left out, a default "sequences.fasta" file is generated which happens to match the `output.sequences` parameter. When modifying  the `output.sequences` to a different filename, the pipeline would fail.

This bug surfaced while writing the dengue ingest where output.sequence needed to be parameterized across serotypes (e.g. "sequences_denv1.fasta", "sequences_denv2.fasta"). 

### Testing

- [ ] Checks pass


joverlee521 (approved):
Thank you for following up on this! (I totally forgot 🤦‍♀️ )

PR #154, opened 2023-04-24
[wrangle metadata] Uses csvtk instead of a separate script

j23414:

Uses csvtk instead of a separate wrangle script as discussed in this thread:

https://github.com/nextstrain/zika/pull/25#discussion_r1117734453

### Related issue(s)

* https://github.com/nextstrain/zika/pull/25#discussion_r1117734453
* https://github.com/nextstrain/dengue/pull/6/commits/f324f443a6e10922613829c5e83b7ab5b4394977

### Testing

- [x] Checks pass


workflow/snakemake_rules/core.smk
```diff
@@ -22,9 +22,8 @@ rule wrangle_metadata:
         strain_id=lambda w: config.get("strain_id_field", "strain"),
     shell:
         """
-        python3 scripts/wrangle_metadata.py --metadata {input.metadata} \
-                    --strain-id {params.strain_id} \
-                    --output {output.metadata}
+        csvtk -t rename -f strain -n strain_original {input.metadata} \
```
victorlin, line 25: Suggestion: explicitly mention the [differences](https://github.com/nextstrain/zika/pull/25#discussion_r1117734453) in the commit message. Otherwise I'd think it's a direct replacement with no functional differences.

workflow/snakemake_rules/core.smk
victorlin, line 1: I don't see any evidence of testing - is this covered by CI?

j23414: I see a CI/pathogen-ci test, which I believe processes [example_data](https://github.com/nextstrain/monkeypox/tree/master/example_data)

* CI Code: [pathogen-repo-ci.yaml#L135-L144](https://github.com/nextstrain/.github/blob/2eb4a5423781c5fdf03c6e1cd8c98daf1e217745/.github/workflows/pathogen-repo-ci.yaml#L135-L144)
* Run Results: [monkeypox/actions/runs/4788819271/jobs/8515926283](https://github.com/nextstrain/monkeypox/actions/runs/4788819271/jobs/8515926283?pr=154)

Ah, are you referring to a different testing framework? Maybe something more modular?

workflow/snakemake_rules/core.smk
j23414: Agree, will write up the differences and add to commit message.

workflow/snakemake_rules/core.smk
victorlin: Yes, I'm referring to pathogen-ci, but am wondering if the successful run is enough to say "ok this change works".

I see it produces a `results/metadata.tsv` file, but is there a way to verify that the contents are as expected?

j23414: Fair point, manually I can verify that the contents are as expected using the following commands:

```
#! /usr/bin/env bash

# Download files metadata.tsv and the wrangle_metadata.py script
wget https://raw.githubusercontent.com/nextstrain/monkeypox/master/example_data/metadata.tsv
wget https://raw.githubusercontent.com/nextstrain/monkeypox/master/scripts/wrangle_metadata.py

# Set variables, normally these would be passed in as parameters
input_metadata='metadata.tsv'
params_strain_id='accession'
output_metadata1='wrangled_metadata_original.tsv'
output_metadata2='wrangled_metadata_new.tsv'

# Original
python3 wrangle_metadata.py --metadata ${input_metadata} \
                    --strain-id ${params_strain_id} \
                    --output ${output_metadata1}

# New
csvtk -t rename -f strain -n strain_original ${input_metadata} \
            | csvtk -t mutate -f ${params_strain_id} -n strain > ${output_metadata2}

# Column order is different, so we need to reorder them before comparison
FIELDS=`csvtk headers -t ${output_metadata1} | tr '\n' ','| sed 's/,$//'`
csvtk cut -t -f $FIELDS wrangled_metadata_new.tsv > new.tsv

# Diff the two files
diff wrangled_metadata_original.tsv new.tsv 
# Nothing should print as different
```

Agree that the above commands should be provided in the PR description to ease the review process.

But having a more fine-grained CI system where we compare against expected intermediate files (maybe within an "example_results" folder) might help ensure equivalent functionality (and also document edge-cases) within these kinds of commits.

workflow/snakemake_rules/core.smk
j23414: Finished editing commit messages, welcome any suggestions on wording/scope

workflow/snakemake_rules/core.smk
victorlin, line 1: From 29080884d6fcd266e7f2e862fe53d2bea6ae5440:

> Unlike the wrangle script which ensures that the "strain" column is the
first column, csvtk appends 'strain' as the last column. Since order of metadata
columns should not affect workflows, this is fine.

I'll note that `augur.io.metadata.read_table_to_dict` (used by `augur curate`) will use the first column as the ID column if unspecified, so the order matters if that function is being called. I don't think it matters in this case though.

workflow/snakemake_rules/core.smk
```diff
@@ -22,9 +22,8 @@ rule wrangle_metadata:
...
+            | csvtk -t mutate -f {params.strain_id} -n strain > {output.metadata}
```
victorlin, line 26: _(non-blocking)_

```suggestion
            | csvtk mutate -t -f {params.strain_id} -n strain > {output.metadata}
```

At first, `csvtk -t mutate` read to me as if `-t` was given the `mutate` value and I wondered what that meant. Only after looking at the [docs](https://bioinf.shenwei.me/csvtk/usage/#mutate) did I understand that `mutate` is a subcommand. Those docs are inconsistent though – both `csvtk -t -H filter` and `csvtk filter2 -H -t` are used for examples.

victorlin (approved):

workflow/snakemake_rules/core.smk
corneliusroemer: Thanks @j23414! This is a neat replacement of a custom script with a one-liner.

I'd just run the full workflow and see it works as intended for testing - pathogen-ci isn't enough to give confidence.

It's very easy to test manually:
```
nextstrain build --docker --cpus 5 . --configfile config/config_mpxv.yaml
```

Looks good in my test, if there's an issue we can fix in production - monkeypox is not a very busy repo.

I don't think we have easy ability to run branches through Github actions here, that's something we could add at some point.

corneliusroemer (approved):

PR #157, opened 2023-06-09
fix: annotations is a file not a param

j23414:

I was wondering why the dot was needed in 

```
  annotations: './source-data/annotations.tsv'
```

and not in 

```
  local_geolocation_rules: 'source-data/geolocation-rules.tsv'
```

The answer is that 'local_geolocation_rules" were marked as [input files in the snakemake rule](https://github.com/nextstrain/monkeypox/blob/46fc9acc0eadb6fe9e7c8db77245f8e4d857a224/ingest/workflow/snakemake_rules/transform.smk#L30). 

This change marks the annotation as an input file, basically the real fix of the warning mentioned in  https://github.com/nextstrain/monkeypox/pull/153 

### Testing

- [ ] Checks pass

### DIFF
--- a/ingest/config/config.yaml
+++ b/ingest/config/config.yaml
@@ -41,7 +41,7 @@ transform:
   # Local rules can overwrite the general geolocation rules provided above
   local_geolocation_rules: 'source-data/geolocation-rules.tsv'
   # User annotations file
-  annotations: './source-data/annotations.tsv'
+  annotations: 'source-data/annotations.tsv'
   # ID field used to merge annotations
   annotations_id: 'accession'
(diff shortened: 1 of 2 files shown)


j23414:
Manual check seems to pass.

```
git clone https://github.com/nextstrain/monkeypox.git original
git clone https://github.com/nextstrain/monkeypox.git changed

cd original/ingest
nextstrain build . 

cd ../../changed/ingest
git checkout fix_annotations_path
nextstrain build . 

diff data/metadata.tsv ../../original/ingest/data/metadata.tsv | less # no changes
```

victorlin (approved):

PR #159, opened 2023-07-06
Switch automation to using the new --env option of `nextstrain build`

tsibley:
Obsoletes the need to write an envdir and use roundabout --exec env + `envdir env.d snakemake` invocations, which in turns obsoletes the need to pass --cores to Snakemake ourselves (and keep it in sync with --cpus).

### Testing

- [ ] Workflows still work
- [ ] Checks pass


tsibley:
I [ran `fetch-and-ingest-branch.yaml`](https://github.com/nextstrain/monkeypox/actions/runs/5480556469/jobs/9983868142) which worked. Testing the others will modify production resources, so I'm not going to before merge.

joverlee521 (approved):
Hooray for this new feature in the Nextstrain CLI 🎉

PR #162, opened 2023-07-28
Use nextstrain/ingest (git subtree)

victorlin:

Begin usage of centralized ingest scripts and add details on how to pull new updates from the central repo.

### Testing

- [x] Ingest workflow runs locally without errors
- [ ] Test with `upload`/`send_slack_notifications`/`trigger_rebuild`?


ingest/README.md
```diff
@@ -85,3 +85,16 @@ These are optional environment variables used in our automated pipeline for prov
 
 GenBank sequences and metadata are fetched via NCBI Virus.
 The exact URL used to fetch data is constructed in `bin/genbank-url`.
+
+## `ingest/vendored`
+
+This repository uses `git subtree` to manage copies of ingest scripts in `ingest/vendored`, from [nextstrain/ingest](https://github.com/nextstrain/ingest). To pull new changes from the central ingest repository, run:
```
victorlin, line 91: Hmm... so `git subtree` (without squash) includes all the commit messages. I had [thought](https://github.com/nextstrain/ingest/issues/3#issuecomment-1656472734) this would be useful, but did not realize that it doesn't play nicely with GitHub's reference-linking feature. See for example the reference to `#6` in 9c9acd631ec65f54b8dff3ebf8ce687119817295 which pollutes the #6 in this repo with a reference, when it's really a reference to nextstrain/ingest#6.

It's pretty obvious that every numerical issue/PR reference in nextstrain/ingest's git history will add a misleading reference in this repo. This is unfortunate, and can be avoided by using `git subtree --squash` or `git subrepo` if we decide to reconsider those options.

joverlee521: I'm in favor of using `git subtree --squash` to keep git history clean. I'm not sure I see the need to see the ingest commit messages in the pathogen repos. I think I would rather direct maintainers to the source ingest repo to see the full history.

PR #164, opened 2023-07-31
Use nextstrain/ingest

victorlin:
### Description of proposed changes

Begin usage of centralized ingest scripts and add details on how to pull new updates from the central repo.

### Testing

- [x] Ingest workflow runs locally without errors
- [x] Test with `upload` ([ref](https://github.com/nextstrain/monkeypox/pull/164#issuecomment-1659264789))

Post-merge:

- [ ] Slack notifications from the next run are as expected.

PR #166, opened 2023-08-04
Use centralized `apply-geolocation-rules`

victorlin:

There have been a few updates to nextstrain/ingest since #164 was merged. I pulled those in with `git subrepo`, and swapped over to use the one newly added script.

### Related issue(s)

Follow-up to #164.

### Testing

- [x] Ingest workflow runs locally without errors


jameshadfield (approved):

ingest/vendored/apply-geolocation-rules
victorlin, line 1: Ingest failed locally with the following error:

```
/bin/bash: ./vendored/apply-geolocation-rules: /usr/bin/env: bad interpreter: Permission denied
```

This is because I didn't add the executable flag on the script in https://github.com/nextstrain/ingest/pull/4. Will fix in that repo then update this PR.

PR #167, opened 2023-08-08
Use centralized scripts with S3 interactions

victorlin:
**Note**: The "Files changed" tab doesn't provide a useful diff since the vendored scripts were added as part of #166. 79e690f465cfafc68157c18b499daf10c36b648d...02d6119a73b1faacbe0879456f47c7ef18a33edf is a more useful diff that detects the script replacements as renames.

Updating to use new scripts added to central ingest in https://github.com/nextstrain/ingest/pull/12.

### Related issue(s)

- Follow-up to #164

### Testing

- [x] Checks pass (failures are [unrelated](https://www.githubstatus.com/incidents/nwb8m3ppfxw6))
- [x] ~Ingest runs locally~ Skipping as this wouldn't test any of the changes
- [ ] Post-merge: Slack notifications in `#monkeypox-updates` are normal


ingest/workflow/snakemake_rules/trigger_rebuild.smk
```diff
@@ -15,5 +15,5 @@ rule trigger_build:
         touch("data/trigger/rebuild.done")
     shell:
         """
-        ./bin/trigger-on-new-data {input.metadata_upload} {input.fasta_upload}
+        ./vendored/trigger-on-new-data nextstrain/monkeypox rebuild {input.metadata_upload} {input.fasta_upload}
```
joverlee521, line 18: Ah, this made me realize `trigger-on-new-data` would need the repo name (without org) since it [calls on the centralized trigger](https://github.com/nextstrain/ingest/blob/5d908187d13cbce27253af5163b4803d6bac03a6/trigger-on-new-data#L22) that [only uses the repo name](https://github.com/nextstrain/ingest/blob/5d908187d13cbce27253af5163b4803d6bac03a6/trigger#L45).

I think it would be preferable to change `trigger` to accept org/repo to match our `notify-*` scripts?

victorlin: Oh good catch, changing to org/repo for consistency makes sense! Are you able to do that in nextstrain/ingest? I can update this PR once that's done.

joverlee521: Ah, thanks for updating nextstrain/ingest!

victorlin: Done in [force-pushes](https://github.com/nextstrain/monkeypox/compare/bf3c93ae25486de943f512fd9b9ef309059168bd..02d6119a73b1faacbe0879456f47c7ef18a33edf).

joverlee521 (approved):

PR #169, opened 2023-08-09
Rename monkeypox -> mpox (virus) when not affecting code

corneliusroemer:
Further changes left over from #165

This PR just changes text that isn't used by code, so shouldn't break anything.


ingest/bin/notify-on-record-change
victorlin, line 1: Let's wait until #167 is merged, which would remove the need to change this file.

PR #171, opened 2023-08-10
Add B.1 build, downsample B.1 in non-B.1 builds, rename to mpox

corneliusroemer:
New build names/URLs no longer include the deprecated `monkeypox` name

New paths are as follows (this can be discussed):
- `monkeypox/mpxv` -> `mpox/mpxv`
- `monkeypox/hmpxv1` -> `mpox/mpxv-IIb` (use official `clade IIb` in name, instead of unofficial `hmpxv1`)

Add new build:
- `mpox/mpxv-IIb-B.1` which is purely B.1 and sub-lineages

Change subsampling of the two non-B.1 builds to include only relatively few B.1s so as to allow examination of non-B.1 diversity. Those who want to study B.1 should use that dedicated build instead.

Sampling for the two non-B.1 builds is so that (almost) all available good quality non-B.1 sequences are included.

The B.1 build includes almost all B.1 sequences.

This is how the IIb build looks like for example:
image

### Testing

- [x] Test builds:
  - [x] mpxv: https://next.nextstrain.org/staging/trial/2/mpox/mpxv
  - [x] mpxv-IIb: https://next.nextstrain.org/staging/trial/2/mpox/mpxv-IIb
  - [x] mpxv-IIb-B.1: https://next.nextstrain.org/staging/trial/2/mpox/mpxv-IIb-B.1

### Follow up

- [ ] Change nextstrain.org manifest to point to new build urls


rneher:
Thanks, Cornelius. This looks very good to me!

rneher (approved):

config/config_hmpxv1.yaml
```diff
@@ -15,21 +15,55 @@ strain_id_field: "accession"
...
-filters: "--exclude-where outbreak!=hMPXV-1"
+
+
+### Set 1: Non-B.1 sequences: use all
+### Set 2: B.1 sequences: small sample across year/country, maybe month
+filter:
+  non_b1:
+    group_by: "--group-by lineage year country"
+    sequences_per_group: "--sequences-per-group 50"
+    other_filters: "outbreak!=hMPXV-1 clade!=IIb"
```
victorlin, line 31: CI is failing because the example data uses old clades. Can you [update](https://github.com/nextstrain/monkeypox/blob/cb7701f7acbc7fd35e98bb6c06668329b30ff870/README.md?plain=1#L121-L126)?

victorlin (changes requested):

config/config_hmpxv1.yaml
corneliusroemer: Done, thanks!

victorlin:
The "commits" and "files changed" tabs are contain unrelated commits/diffs, I think due to the way the rebase was applied. This should be more reflective of actual changes from this PR: 942c1d00dd163ed837062828ad2869be2822a9ac...c735b6adecc462890f54269ac0d425757d84988f

PR #173, opened 2023-08-22
Use default config and partial overrides

victorlin:

Refactoring the way Snakemake configuration is handled in this workflow to be more consistent with the ncov workflow.

### Related issue(s)

- Prompted by https://github.com/nextstrain/docs.nextstrain.org/pull/164#discussion_r1294800635

### Testing

- [x] Checks pass

### DIFF
--- a/config/defaults.yaml
+++ b/config/defaults.yaml
@@ -1,0 +1,29 @@
+# This configuration file contains configuration values that should apply to all
+# builds of this pathogen repo.
+# Some required keys are missing and must be additionally defined through
+# Snakemake's --configfile/--config options. Those options can also override
+# any values in this file.
+
+exclude: "config/exclude_accessions_mpxv.txt"
+reference: "config/reference.fasta"
+genemap: "config/genemap.gff"
+genbank_reference: "config/reference.gb"
+clades: "config/clades.tsv"
+lat_longs: "config/lat_longs.tsv"
+description: "config/description.md"
+tree_mask: "config/tree_mask.tsv"
+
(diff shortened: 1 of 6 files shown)


config/defaults.yaml
```diff
@@ -1,15 +1,15 @@
...
-# any values in this file.
+# This configuration file contains all required configuration entries with
+# hmpxv1-focused default values.
+# Use Snakemake's --configfile/--config options to override default values.
 
 exclude: "config/exclude_accessions_mpxv.txt"
 reference: "config/reference.fasta"
 genemap: "config/genemap.gff"
 genbank_reference: "config/reference.gb"
+colors: "config/colors_hmpxv1.tsv"
```
joverlee521, line 9: _nitpick_

I can't find where this config is being accessed in the workflow.
Maybe this should have been removed as part of https://github.com/nextstrain/monkeypox/pull/109?

joverlee521 (approved):
Thanks for making this change! Having a reasonable default config file is definitely a change I want to push across all Nextstrain pathogen repos!

config/defaults.yaml
victorlin: Hmm, I don't see any references either. I'll make another PR to remove it.

victorlin: #174

corneliusroemer:
Hmm, I don't like having an extra default config file which makes it more difficult to trace where a particular config value comes from.

PR #171 reworks things a little anyways. I'll add a default profile which specifies a default config file so things can be run as `snakemake`.

PR #175, opened 2023-08-30
Use centralized scripts for NCBI Virus

victorlin:

`fetch-from-ncbi-virus` from the subrepo is a copy of `fetch-from-genbank` with some extra options. It can be used as-is since the new options are not required. Remove the copy in this repo and update references.

The supporting scripts `csv-to-ndjson` and `genbank-url` can also be removed.

### Related issue(s)

- Follow-up to https://github.com/nextstrain/ingest/pull/16
- Follow-up to https://github.com/nextstrain/monkeypox/pull/164

### Testing

- [x] Checks pass (failure is [unrelated](https://github.com/nextstrain/monkeypox/issues/177))
- [x] `nextstrain build --cpus 1 . data/sequences.ndjson` runs locally


ingest/vendored/fetch-from-ncbi-virus
victorlin, line 1: I think this works on Bash 5.0, however it failed on my Mac running Bash 3.7. I'm fixing in https://github.com/nextstrain/ingest/pull/17.

joverlee521 (approved):

PR #179, opened 2023-09-11
ingest: Switch to NCBI Datasets CLI to fetch data

### LINKED ISSUE #178: Ingest currently blocked by `fetch-from-ncbi-virus`
### Current Behavior
Because of the behavior described in https://github.com/nextstrain/ingest/issues/18, the ingest pipeline does not include sequences in it's fetch from NCBI Virus. This results in all of the records being dropped in the pipeline and the final outputs to s3://nextstrain-data/files/workflows/monkeypox/ are empty. This was first flagged internally by downstream CZI consumers on [Slack](https://bedfordlab.slack.com/archives/C021ZG2P8NN/p1694445677860929). 

We don't have insight into the undocumented NCBI Virus API and whether this new behavior is intentional, so the best thing might be to just switch to the NCBI Datasets CLI to fetch data.

joverlee521:
Replace our custom fetch scripts that uses the NCBI Virus API with the NCBI Datasets CLI commands.

NCBI datasets downloads a virus dataset ZIP file that includes a metadata JSON Lines file and a sequences FASTA file. To maintain a record of the single NDJSON file on S3, extract the sequences FASTA file and format the metadata into a TSV file that are parsed into a single NDJSON file using `augur curate passthru`. The metadata TSV is created using the NCBI `dataformat` command so that we do not have to parse the nested JSON lines files ourselves and header fields are renamed to match the previous fields we used for NCBI Virus.

The NDJSON file created here no longer includes equivalent fields for "title" or "publication".

## Related issue(s)

Resolves #178 

## Checklist

- [x] Checks pass
- [x] [Test ingest run](https://github.com/nextstrain/monkeypox/actions/runs/6152797615)
- [x] Diff test run outputs with previous versions.

### DIFF
--- a/ingest/source-data/ncbi-dataset-field-map.tsv
+++ b/ingest/source-data/ncbi-dataset-field-map.tsv
@@ -1,0 +1,17 @@
+key	value
+Accession	genbank_accession_rev
+Source database	database
+Isolate Lineage	strain
+Geographic Region	region
+Geographic Location	location
+Isolate Collection date	collected
+Release date	submitted
+Update date	updated
+Length	length
+Host Name	host
+Isolate Lineage source	isolation_source
+BioProjects	bioproject_accession
+BioSample accession	biosample_accession
+SRA Accessions	sra_accession
+Submitter Names	authors
+Submitter Affiliation	submitting_organization
--- a/ingest/workflow/snakemake_rules/fetch_sequences.smk
+++ b/ingest/workflow/snakemake_rules/fetch_sequences.smk
@@ -16,8 +16,8 @@ Produces final output as
 
 rule fetch_ncbi_dataset_package:
     output:
-        dataset_package = temp("data/ncbi_dataset.zip")
-    retries: 5 # Requires snakemake 7.7.0 or later
+        dataset_package=temp("data/ncbi_dataset.zip"),
+    retries: 5  # Requires snakemake 7.7.0 or later
     benchmark:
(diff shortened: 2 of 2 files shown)


joverlee521:
Compared `s3://nextstrain-data/files/workflows/monkeypox/branch/ncbi-datasets-cli/metadata.tsv.gz` with `s3://nextstrain-data/files/workflows/monkeypox/metadata.tsv.gz` (version id `WYEDFJJLxGxrbB.hmWkC2noCfRzrX1Ky`). 

There were 2 records removed ([8HG1_P](https://www.ncbi.nlm.nih.gov/nuccore/8HG1_P) and [8HG1_T](https://www.ncbi.nlm.nih.gov/nuccore/8HG1_T)), 90 records added, and 6340 records changed. 
Of the changed records:

- 6333 had white space changes in the authors column 
  (e.g. `Babkin,I.V., Babkina,I.N., Shchelkunov,S.N.->Babkin,I.V.,Babkina,I.N.,Shchelkunov,S.N.`)

- 16 had quotation changes in the institution column
  (e.g. `"Babkin I.V., Molecular Biology of Genoms, SRC VB ""VECTOR"", Koltcovo, Novosibirsk region, 633159->Babkin I.V., Molecular Biology of Genoms, SRC VB ""VECTOR"""", Koltcovo, Novosibirsk region, 633159"""`)

The quotation changes come from the extra pass through `augur curate passthru`. We should fix this behavior in Augur, but I'm not going to let that block this PR.

PR #185, opened 2023-09-21
Reorganize config and rules around filter/subsampling

victorlin:

Subsampling is a form of filtering, but there is a conceptual distinction that is apparent in the rules and configs used in the ncov workflow, which I've mirrored here.

Put all filtering-related rules in a "filter" config, which configures a rule with the same name.

Put all subsampling-related rules in a "subsampling" config which also configures a rule with the same name.

## Related issue(s)

Follow-up to #171

## Checklist

- [x] Checks pass


corneliusroemer (approved):

corneliusroemer:
Thanks, that makes sense!

PR #188, opened 2023-09-22
Move phylogenetic workflow to folder

corneliusroemer:
As this repo will contain 3 separate workflows soon (ingest, nextstrain builds, nextclade datasets) it makes sense to move the nextstrain build workflow to its own folder so that the top level is nice and clean.


phylogenetics/README.md
victorlin, line 1: `phylogenetics` or `phylogenetic`?

I'm assuming this will mirror @joverlee521's [pathogen-repo-template](https://github.com/nextstrain/pathogen-repo-template), which currently uses `phylogenetic`.

corneliusroemer: `phylogenetic` is an adjective which is just hanging there, begging for a noun whereas  `phylogenetics` can stand for free. I don't know what the reasoning was to choose `phylogenetic` for the template. I guess my proposal would be to change it in the template.

corneliusroemer:
This is a draft, I'll redo it with a force-push as there've been changes to README etc that would be annoying to resolve as merge conflicts.

corneliusroemer:
Close in favor of #198

PR #189, opened 2023-09-22
ingest: Add `profiles/default` to make `-c1 --ri` unnecessary

corneliusroemer:


ingest/profiles/default/config.yaml
```diff
@@ -0,0 +1,4 @@
+cores: all
+rerun-incomplete: true
+printshellcmds: true
+reason: true
```
victorlin, line 4: These options are newly introduced, but seems reasonable to me.

victorlin (approved):

ingest/profiles/default/config.yaml
```diff
@@ -0,0 +1,4 @@
+cores: all
```
victorlin, line 1: Changed from `1` to `all` but I think that's fine.

corneliusroemer: I don't know why we ever used 1 in the first place

ingest/profiles/default/config.yaml
corneliusroemer: Yep, rerun incomplete just avoids an error, it's something almost everyone adds all the time anyways when running builds

PR #190, opened 2023-09-22
Use pathogen-repo-build workflow for builds

corneliusroemer:
_Original commit is from @joverlee521, I just rebased it on master_

Simplify build workflows by using the reusable Nextstrain pathogen-repo-build workflow.

This does not update the fetch-and-ingest-branch workflow since that will require additional changes to support the custom configs.

This does remove the `notify_pipeline_failed` step in the workflows. They can be added as a separate job in the workflow if we think they are still helpful.

### DIFF
--- a/.github/workflows/fetch-and-ingest.yaml
+++ b/.github/workflows/fetch-and-ingest.yaml
@@ -31,27 +31,20 @@ on:
         description: 'Specific container image to use for build (will override the default of "nextstrain build")'
         required: false
 
-env:
-  NEXTSTRAIN_DOCKER_IMAGE: ${{ github.event.inputs.image }}
-
 jobs:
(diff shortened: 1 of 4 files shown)


corneliusroemer:
IIUC, `notify_pipeline_failed` _only_ notified when the _Github action_ failed, not when the build itself failed. Am I right @joverlee521?

corneliusroemer:
This looks good to me, I'll merge and bear in mind if it fails

joverlee521:
> IIUC, notify_pipeline_failed only notified when the Github action failed, not when the build itself failed. Am I right @joverlee521?

Yup, this was added to catch errors that were raised outside of the build itself. We should still get Slack notifications if there is an error _during_ the build.

PR #187, opened 2023-09-22
chore: apply snakefmt

corneliusroemer:

PR #193, opened 2023-09-22
Add pyproject.toml for sane `snakefmt` excludes, add `.git/` to `.gitignore`

corneliusroemer:
See https://github.com/nextstrain/monkeypox/pull/187 for details
And https://github.com/snakemake/snakefmt/issues/202

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,10 @@ environment*
 env.d/
 
 # Snakemake state dir
-.snakemake
+.snakemake/
+
+# Git dir
+.git/
 
 # Local config overrides
 /config_local.yaml
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[tool.snakefmt]
+include = '\.smk$|^Snakefile'
+exclude = ''


.gitignore
```diff
@@ -19,7 +19,10 @@ environment*
 env.d/
 
 # Snakemake state dir
-.snakemake
+.snakemake/
+
+# Git dir
+.git/
```
victorlin, line 25: Did you run into a situation where `.git` was not ignored by git locally? I've never seen that before and I'd think git should know to ignore its own internals...

corneliusroemer: Git knows how to ignore git but snakefmt doesn't - that's why I added it. Not necessary as nothing within `.git` will end in `.smk` but just in case

victorlin: Oh I see. `snakefmt` doesn't read from `.gitignore` though, so this has no effect. Instead, it should be added in `snakefmt`'s exclude config.

victorlin: Scratch that, [looks like it's supposed to](https://github.com/snakemake/snakefmt/blob/618df85a5b16dd8f6daef1b30e4a6a38d76ac8b1/snakefmt/snakefmt.py#L52-L54). I'm not sure what's going on locally then...

corneliusroemer: My thinking is: let's keep the exclude as light as possible (empty if possible). Putting the `.git` in `.gitignore` ensures that snakefmt doesn't run on things it shouldn't run on in any case.

PR #196, opened 2023-09-23
Add yamlfmt to pre-commit (and format with it)

corneliusroemer:


corneliusroemer:
Added in separate commit

PR #198, opened 2023-09-25
Move phylogenetic workflow from top-level to folder `phylogenetic`

corneliusroemer:
Decided to use `phylogenetic`


.github/workflows/ci.yaml
```diff
@@ -9,7 +9,10 @@ on:
 
 jobs:
   pathogen-ci:
-    uses: nextstrain/.github/.github/workflows/pathogen-repo-ci.yaml@master
+    uses: nextstrain/.github/.github/workflows/pathogen-repo-ci.yaml@dec0880059017dac7facf100435c5737bf1386c8
+    with:
+      workflow-root: phylogenetic
```
tsibley, line 14: This is still using the version which was reverted.

corneliusroemer: Thanks, yes I'm aware, worst case is CI fails but as this is under active development right now that's ok. Alternative is to disable pathogen-repo-ci and roll it on its own, but that's not worth the effort

README.md
```diff
@@ -1,106 +1,40 @@
...
-```
+This repository uses [pre-commit](https://pre-commit.com/) to run checks on the code before committing.
 
-Run pipeline to produce the "lineage B.1" tree for `/mpox/lineage-B.1` with:
+To install pre-commit on macOS, run:
 
 ```bash
-nextstrain build . --configfile config/hmpxv1_big/config.yaml
+brew install pre-commit
 ```
```
victorlin, line 24: Is this a requirement for anyone working on this repo? What happens if I don't install pre-commit and push something to `master` that would fail checks?

corneliusroemer: It's just like snakefmt lint checks, nothing happens other than that pre-commit checks fail and the commit gets a red ❌

PR #200, opened 2023-09-26
general precommit lints

corneliusroemer:
- Add general precommit lints
- Add codespell to precommit
- Add shellcheck to precommit
- Add tomlsort to precommit

## Checklist

- [ ] Checks pass


.pre-commit-config.yaml
victorlin, line 1: This file and pre-commit are new to me. I think this is what happened:

1. This config file was added outside this PR in 208e26a41550e02b234cf49b91fd28d9ea12cfbb.
2. pre-commit was added [as a GitHub app](https://github.com/organizations/nextstrain/settings/installations/42164746) with limited write access to this repository.
3. Further instructions in README were added as a part of 03f9a253b23f67336940655bb280b0fa17d90685

@corneliusroemer did you do anything else to configure this? It would be good to document such things at least in the PR description. Something like [this](https://github.com/nextstrain/nextstrain.org/pull/667):

> ### Changes outside of this PR
> 
> - [x] Regenerated `[org secret] GH_TOKEN_NEXTSTRAIN_BOT_REPO` and renamed as `[org Actions+Dependabot secret] GH_TOKEN_NEXTSTRAIN_BOT_REPO`
> - [x] Created a new organization Dependabot secret on named `GH_TOKEN_NEXTSTRAIN_BOT_REPO` with the regenerated value
> - [x] Updated the value of the existing organization Actions secret with the same name
> - [x] Added `dependabot/fetch-metadata@*` as an allowed GitHub action on @nextstrain.

README.md
victorlin, line 1: Reviewing related README changes here: https://github.com/nextstrain/monkeypox/pull/198#discussion_r1337572426

.pre-commit-config.yaml
corneliusroemer: Yes, there's a config file that configures the checks to run

You can run precommit locally and/or in CI. It's particularly useful as it allows you to run checks before things are committed or pushed, but this is not a necessary requirement. Same for CI, one can also not use the app and implement via a Github action (deprecated/discouraged) or not run in CI at all.

Nothing breaks if you don't use it, it's like a formatter, just makes it easy to keep things consistent.

PR #201, opened 2023-09-27
fix: correct vendored path in notify-on-* scripts after move and build action args

corneliusroemer:
#### Vendored scripts after move at different relative path

When moving the workflow to folder phylogenetic
we need to adjust the path to the vendored script dir
which is used by Slack notify-on* scripts in the phylogenetic
workflow

If we ever centralize these scripts, we might want to allow
the vendor dir to be passed via an argument
this was quite a fixed assumption deep in the code, hard to
find before testing

See changes: https://github.com/nextstrain/monkeypox/pull/201/files#diff-b30e0f3bd03a03bd7a30bde82b5a2254893ea8369c841f79c3c9168b9136a18e

#### `pathogen-repo-build` args need to be adjusted to account for fact that files above the workflow root are needed

See changes: https://github.com/nextstrain/monkeypox/pull/201/files#diff-e41bb7e56e6e234a06e78585e851d80da70f1a0baea79f6f507bbe9814a23566

### Test
- [x] https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#jobs/detail/179eb2a3-546f-4e33-90ae-4a13aa26412f

### DIFF
--- a/.github/workflows/rebuild-mpxv.yaml
+++ b/.github/workflows/rebuild-mpxv.yaml
@@ -47,4 +47,4 @@ jobs:
           . \
             notify_on_deploy \
               --configfiles $BUILD_DIR/config/mpxv/config.yaml $BUILD_DIR/config/nextstrain_automation.yaml \
-              --config auspice_prefix=$TRIAL_NAME --directory $BUILD_DIR
+              --config auspice_prefix=$TRIAL_NAME --directory $BUILD_DIR --snakefile $BUILD_DIR/Snakefile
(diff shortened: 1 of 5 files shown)


j23414:
Thanks @corneliusroemer! 

Does make me wonder if we should revisit `workflow.current_basedir ` (https://github.com/nextstrain/dengue/pull/10/commits/df024cb9462a7b85396963f698a86bed74bb25ab)

corneliusroemer:
Oof, that looks complicated, I've looked at the commit and have no clue what's going on there. Modules sound like a whole new level of complexity that's best avoided? @j23414 :)

joverlee521:
I tried adding the [Snakemake modules to this repo](https://github.com/nextstrain/monkeypox/compare/master...top-level-module) but I think it might be more trouble than it's worth. 

The main pain point for me is trying to figure out which files/scripts would need to be prefixed with  `workflow.current_basedir`. I hadn't realized that file paths provided via config parameters would also need to be prefixed, so it's not as simple as only prefixing local shell scripts.  

It's probably more straight-forward to do what @corneliusroemer has done in this PR by providing the `--directory` and `--snakefile` options to run the subworkflows.

PR #206, opened 2023-10-03
Remove extra actionlint

joverlee521:

Noticed the extra actionlint when I was catching up on the addition of pre-commit to this repo. 
Remove extra actionlint and sort pre-commit config by repo name to avoid duplicates in the future. 

## Checklist

- [ ] Checks pass


joverlee521:
Changes order in which the hooks are run in pre-commit CI, but order shouldn't really matter here. All hooks are run with the default `False` value for the `fail_fast` option, so all hooks are run regardless of failure status of other hooks.

victorlin (approved):

PR #207, opened 2023-10-04
Replace join metadata and clades script with csvtk and tsv append

j23414:

In our effort to centralize ingest scripts, we identified an opportunity to streamline the process by replacing the `join-metadata-and-clades.py` script with `csvtk rename` and `tsv-append` in cases where customized score calculations are not required.

The specific columns from a nextclade TSV file to be renamed (if desired) and appended to the final metadata.tsv file are defined in a `source-data/nextclade-field-map.tsv` key-value file. This implementation is similar to the approach used in [ncbi-source-field-map](https://github.com/nextstrain/monkeypox/blob/1f169d2b5c9ef28a870816282f6ee51d037fd3f0/ingest/workflow/snakemake_rules/fetch_sequences.smk#L93) used elsewhere.

## Related issue(s)

https://github.com/nextstrain/ingest/pull/23

## Checklist

- [x] Checks pass
- [x] Manual check

Manual check could look like: 

```
git clone https://github.com/nextstrain/monkeypox.git old
git clone https://github.com/nextstrain/monkeypox.git new

cd new/ingest
git checkout replace_join_metadata_and_clades
nextstrain build . data/metadata.tsv

cd ../../old/ingest
nextstrain build . data/metadata.tsv

# Compare using daff diff since order of columns were not retained
daff diff data/metadata.tsv ../../new/ingest/data/metadata.tsv > diff.txt
```


ingest/workflow/snakemake_rules/nextclade.smk
```diff
@@ -56,15 +56,31 @@ rule join_metadata_clades:
...
+
+        export APPEND_FIELDS=`awk 'NR>1 {{print $2}}' {input.nextclade_field_map} | grep -v -e "index" -e "seqName" | tr '\n' ',' | sed 's/,\$//g'`
+
+        cat {input.metadata} | tr '\r' '\n' | grep -v "^$" \
+        | tsv-join -H \
+            --filter-file results/nextclade_renamed.tsv \
+            --key-fields seqName \
+            --data-fields {params.id_field} \
+            --append-fields $APPEND_FIELDS \
+            --allow-duplicate-keys \
```
joverlee521, line 83: Curious why we need to allow duplicates in the join? The `seqName` should be the GenBank accession, which should all be unique.

ingest/source-data/nextclade-field-map.tsv
```diff
@@ -0,0 +1,17 @@
+key	value
+index	index
```
joverlee521, line 2: Do we need the `index` column? From what I can see it gets filtered out in the join and is not used as a join key either.

joverlee521 (changes requested):
Thanks for working through this @j23414! This approach makes sense to me but I'd be curious to see what others think. 

I do want to note that this changes the order of columns in the final metadata TSV. That should be _fine_ for our uses of the metadata TSV. I can't think of any downstream Augur commands that depend on column order. 

There's currently an error that is blocking due to the use of the `results` directory and I've added other comments as well.

ingest/workflow/snakemake_rules/nextclade.smk
```diff
@@ -56,15 +56,31 @@ rule join_metadata_clades:
...
+            -f '*' \
+            -p '(.+)' \
+            -r '{{kv}}' \
+            -k {input.nextclade_field_map} \
+            {input.nextclade} \
+            > results/nextclade_renamed.tsv
+
+        export APPEND_FIELDS=`awk 'NR>1 {{print $2}}' {input.nextclade_field_map} | grep -v -e "index" -e "seqName" | tr '\n' ',' | sed 's/,\$//g'`
+
+        tsv-join -H \
```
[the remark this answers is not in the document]
joverlee521: Also discussed in Slack to use `tsv-select` or `csvtk cut` to subset the Nextclade columns so save disk space for the generated `nextclade_renamed.tsv` file.

ingest/workflow/snakemake_rules/nextclade.smk
```diff
@@ -56,15 +56,31 @@ rule join_metadata_clades:
...
-                --nextclade {input.nextclade} \
-                -o {output}
+        csvtk -tl rename2 \
+            -F \
+            -f '*' \
+            -p '(.+)' \
+            -r '{{kv}}' \
+            -k {input.nextclade_field_map} \
+            {input.nextclade} \
+            > results/nextclade_renamed.tsv
```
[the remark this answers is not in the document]
j23414: Ah, fixed in https://github.com/nextstrain/monkeypox/pull/207/commits/1323f1bb7cfec7fd030ed98543f6414ce42fe0ed

ingest/workflow/snakemake_rules/nextclade.smk
```diff
@@ -74,13 +74,13 @@ rule join_metadata_clades:
 
         export APPEND_FIELDS=`awk 'NR>1 {{print $2}}' {input.nextclade_field_map} | grep -v -e "index" -e "seqName" | tr '\n' ',' | sed 's/,\$//g'`
 
-        tsv-join -H \
+        cat {input.metadata} | tr '\r' '\n' | grep -v "^$" \
```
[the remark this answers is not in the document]
j23414: I opted for the `lineterminator='\n'` route https://github.com/nextstrain/monkeypox/pull/207/commits/233457ababe49468b39ff4797668dd2ac197be83

While I acknowledge that the `augur curate passthru` route would be more elegant, the implementation ended up resembling the following:

```
|  augur curate passthru \
  --output-metadata - \
  --output-fasta data/sequences.fasta \
  --output-id-field accession \
  --output-seq-field sequence \
| tsv-select -H \
  --fields accession,genbank_accession_rev,strain,date,region,country,division,location,host,date_submitted,sra_accession,abbr_authors,reverse,authors,institution  \
  > data/metadata_raw.tsv
```

Perhaps we could consider introducing an `--output-metadata-fields` flag, particularly for fields that are absent (e.g. non-existent column `reverse` causes `tsv-select` to error out).

ingest/workflow/snakemake_rules/nextclade.smk
```diff
@@ -56,15 +56,31 @@ rule join_metadata_clades:
...
+        csvtk -tl rename2 \
+            -F \
+            -f '*' \
+            -p '(.+)' \
+            -r '{{kv}}' \
+            -k {input.nextclade_field_map} \
+            {input.nextclade} \
+            > results/nextclade_renamed.tsv
+
+        export APPEND_FIELDS=`awk 'NR>1 {{print $2}}' {input.nextclade_field_map} | grep -v -e "index" -e "seqName" | tr '\n' ',' | sed 's/,\$//g'`
```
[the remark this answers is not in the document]
j23414: Thanks! Fixed in https://github.com/nextstrain/monkeypox/pull/207/commits/c3843e7f434b47a4b44d69ed2a6448f1511265d7

ingest/workflow/snakemake_rules/nextclade.smk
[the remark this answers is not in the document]
j23414: I really liked this suggestion, implemented in https://github.com/nextstrain/monkeypox/pull/207/commits/d0cf7718cd885085adf1626947a24e086eeb1f3a

ingest/workflow/snakemake_rules/nextclade.smk
j23414: Dropped in https://github.com/nextstrain/monkeypox/pull/207/commits/0ff67dfdd4a967f055c22923386de7e05aca1a33
I use a similar `tsv-join` in an unrelated pipeline where my keys are duplicated 😓 but you are correct, seqName here should be unique

ingest/source-data/nextclade-field-map.tsv
j23414: Dropped in https://github.com/nextstrain/monkeypox/pull/207/commits/0ff67dfdd4a967f055c22923386de7e05aca1a33

ingest/workflow/snakemake_rules/nextclade.smk
joverlee521: (Dropped in c21e9f80797bca1e8c4462328cc23bbfd76f0b8a)

ingest/workflow/snakemake_rules/nextclade.smk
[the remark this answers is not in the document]
joverlee521: (100% fine to leave as-is, just commenting for posterity)

Ah, I specifically didn't feel the need to add an `--output-metadata-fields` option because you can chain it with `tsv-select` to subset the columns 😅 

As for the `reverse` column, it seems like we can completely remove this column from the metadata output since Nextclade includes an `is_reverse_complement` column. The phylogenetic builds now use the `is_reverse_complement` column since https://github.com/nextstrain/monkeypox/pull/191. Maybe @corneliusroemer can help confirm here that the `reverse` column is no longer needed.

joverlee521 (changes requested):
Thanks for fixing things up @j23414! I have one final suggestion to parameterize the `--key-fields` name but everything else looks good by inspection. 

Also, not sure if you've seen this in this repo, but you can do a full test run of the ingest workflow with the [Fetch and ingest (on branch) workflow](https://github.com/nextstrain/monkeypox/actions/workflows/fetch-and-ingest-branch.yaml). When you select to run on a branch with this workflow, the ingest pipeline will upload outputs to `s3://nextstrain-data/files/workflows/monkeypox/branch/<branch-name>/`.

j23414:
Thanks @Jover! Sorry for the delay; I've added the nextclade-key-field parameterization (https://github.com/nextstrain/monkeypox/pull/207/commits/cefff3b3405f9e313c67220fdd96245bd43e4d9b).

Thank you for highlighting the `Fetch and ingest (on branch) workflow`! 
I'm encountering an error -  `./bin/set-branch-ingest-config: No such file or directory` with an exit code of 127 when I attempt to run it. Perhaps I'm overlooking something?

https://github.com/nextstrain/monkeypox/actions/runs/6475385391

ingest/workflow/snakemake_rules/nextclade.smk
```diff
@@ -61,6 +61,7 @@ rule join_metadata_clades:
         metadata="data/metadata.tsv",
     params:
         id_field=config["transform"]["id_field"],
+        nextclade_id_field=config["transform"]["nextclade_id_field"],
```
joverlee521, line 64: _non-blocking_

It's slightly weird to me that this param is under `transform` since it is not being used in the transform rule. I think it's reasonable to move this to a top level config key, but that's just me being nit-picky.

joverlee521 (approved):
Looks good to me! I think to good to merge as long as the trial run shows no difference in the output (except column order in the metadata.tsv).

ingest/workflow/snakemake_rules/nextclade.smk
j23414: I think that's a reasonable and quick change. Perhaps even defining a nextclade config section: https://github.com/nextstrain/monkeypox/pull/207/commits/9817e6e51f3d691a27270b155dec7dfd44623738

joverlee521: That's the direction I'm pushing in the [pathogen-repo-template](https://github.com/nextstrain/pathogen-repo-template/pull/7) as well! 

I think it still works well here even though we are technically using [two different Nextclade datasets](https://github.com/nextstrain/monkeypox/blob/3e836176c959803263dd260726d3325e1bf5d0c8/ingest/workflow/snakemake_rules/nextclade.smk#L2-L17).

ingest/workflow/snakemake_rules/nextclade.smk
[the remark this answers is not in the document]
joverlee521: Moving discussion of `reverse` column to a [new issue](https://github.com/nextstrain/monkeypox/issues/209).

PR #208, opened 2023-10-10
fix: script moved to phylogenetic dir

j23414:

Minor path fix, after migration from https://github.com/nextstrain/monkeypox/pull/198 

Surfaced while testing https://github.com/nextstrain/monkeypox/pull/207

## Checklist

- [ ] Checks pass


joverlee521 (approved):

j23414:
Much later, it occurred to me that placing this script in the `ingest` rather than the `phylogenetic` dir might make more logical sense. However, I don't think it's a priority to move the script at the moment.

PR #215, opened 2023-10-27
Move s3_dst config to top level config

j23414:

Nested configs for s3 URLs are challenging to override by the Snakemake `--config` option, as discussed in the following comment:

https://github.com/nextstrain/dengue/pull/13#discussion_r1361131210

This moves the url to the top level as a `s3_dst` config value, and propagate those changes to the relevant Snakefiles. 
It's important to note that I haven't conducted thorough testing, and there may be nuances in the implementation meant by the comment. Suggestions or clarification welcome. 

Additionally, I was unsure regarding whether the `s3_dst` config should be a required value or an optional default ([Snakemake styleguide: config values](https://docs.nextstrain.org/en/latest/reference/snakemake-style-guide.html#access-config-values-appropriately)). 

## Related issue(s)

* https://github.com/nextstrain/dengue/pull/13#discussion_r1361131210

## Checklist

- [ ] Checks pass

### DIFF
--- a/ingest/Snakefile
+++ b/ingest/Snakefile
@@ -32,7 +32,7 @@ def _get_all_targets(wildcards):
             print(
                 f"Skipping file upload for {target!r} because there are duplicate remote file names."
             )
-        elif not params.get("dst"):
+        elif not params.get("s3_dst"):
             print(
                 f"Skipping file upload for {target!r} because the destintion was not defined."
             )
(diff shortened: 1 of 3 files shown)


j23414:
Since we can use: `--config upload='{"s3": {"dst": "NEW URL"}}'`

https://github.com/nextstrain/dengue/pull/13#discussion_r1375081279

This PR is no longer necessary.

PR #217, opened 2023-10-27
Move ncbi taxon id to config

j23414:

Inspired by the following comment:

* https://github.com/nextstrain/dengue/pull/13#discussion_r1373373042

## Related issue(s)

* https://github.com/nextstrain/dengue/pull/13#discussion_r1373373042

## Checklist

- [x] Checks pass

### DIFF
--- a/phylogenetic/config/description.md
+++ b/phylogenetic/config/description.md
@@ -18,7 +18,7 @@ Our bioinformatic processing workflow can be found at [github.com/nextstrain/mon
 
 #### Underlying data
 We curate sequence data and metadata from the [NCBI Datasets command line tools](https://www.ncbi.nlm.nih.gov/datasets/docs/v2/download-and-install/),
-using NCBI Taxonomy ID "10244", as starting point for these analyses.
+using an NCBI Taxonomy ID defined in [ingest/config/config.yaml](https://github.com/nextstrain/monkeypox/blob/master/ingest/config/config.yaml), as starting point for these analyses.
 Curated sequences and metadata are available as flat files at:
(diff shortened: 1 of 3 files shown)


victorlin (approved):

joverlee521 (approved):
Thanks for keeping things up-to-date in monkeypox! Changes look good to me. 

Consider if we should update the [monkeypox description](https://github.com/nextstrain/monkeypox/blob/c31003754f499116a06aee4cba435bad261ea61a/phylogenetic/config/description.md?plain=1#L21) to point to the config file instead of hardcoding the taxon id. I don't think we will ever change the taxon id, but just in case they might get out of sync.

phylogenetic/config/description.md
```diff
@@ -18,7 +18,7 @@ Our bioinformatic processing workflow can be found at [github.com/nextstrain/mon
 
 #### Underlying data
 We curate sequence data and metadata from the [NCBI Datasets command line tools](https://www.ncbi.nlm.nih.gov/datasets/docs/v2/download-and-install/),
-using NCBI Taxonomy ID "10244", as starting point for these analyses.
+using an NCBI Taxonomy ID defined in [ingest/config/config.yaml](https://github.com/nextstrain/monkeypox/blob/master/ingest/config/config.yaml), as starting point for these analyses.
```
j23414, line 21: I realized I could have submitted this as a separate PR. But would this wording be okay? @joverlee521 ?

joverlee521: LGTM!

PR #220, opened 2023-11-01
Rename instances of "monkeypox" to "mpox"

trvrb:
This commit goes through the repo and basically does a find-and-replace of "monkeypox" to "mpox, except for a handful of locations like the #monkeypox-updates Slack notification. 

This shouldn't change appearance of build outputs except for:
- Updating to "Built by nextstrain/mpox"
- Updating page titles to be more descriptive of what each page is showing in terms of lineages / clades
- A bug fix to the page description to link to proper build URLs (currently these go to 404)

This has the side effect that intermediate files will begin to propagate to https://data.nextstrain.org/files/workflows/mpox/sequences.fasta.xz, etc... rather than `/files/workflows/monkeypox`. This may affect users who've been relying on the current URLs. Nothing should break for these users as the old files will continue to exist but they will no longer receive updated data. I think this is an okay trade-off in terms of downstream impacts.

### DIFF
--- a/ingest/vendored/README.md
+++ b/ingest/vendored/README.md
@@ -12,6 +12,9 @@ Some tools may only live here temporarily before finding a permanent home in
 Nextstrain maintained pathogen repos will use [`git subrepo`](https://github.com/ingydotnet/git-subrepo) to vendor ingest scripts.
 (See discussion on this decision in https://github.com/nextstrain/ingest/issues/3)
 
+For a list of Nextstrain repos that are currently using this method, use [this
+GitHub code search](https://github.com/search?type=code&q=org%3Anextstrain+subrepo+%22remote+%3D+https%3A%2F%2Fgithub.com%2Fnextstrain%2Fingest%22).
+
 If you don't already have `git subrepo` installed, follow the [git subrepo installation instructions](https://github.com/ingydotnet/git-subrepo#installation).
 Then add the latest ingest scripts to the pathogen repo by running:
 
@@ -23,6 +26,13 @@ Any future updates of ingest scripts can be pulled in with:
 
 ```
 git subrepo pull ingest/vendored
+```
+
+If you run into merge conflicts and would like to pull in a fresh copy of the
+latest ingest scripts, pull with the `--force` flag:
+
+```
+git subrepo pull ingest/vendored --force
 ```
 
 > **Warning**
@@ -47,14 +57,14 @@ commit hash if needed.
 
 Much of this tooling originated in
 [ncov-ingest](https://github.com/nextstrain/ncov-ingest) and was passaged thru
-[monkeypox's ingest/](https://github.com/nextstrain/monkeypox/tree/@/ingest/).
-It subsequently proliferated from [monkeypox][] to other pathogen repos
-([rsv][], [zika][], [dengue][], [hepatitisB][], [forecasts-ncov][]) primarily
-thru copying.  To [counter that
+[mpox's ingest/](https://github.com/nextstrain/mpox/tree/@/ingest/). It
+subsequently proliferated from [mpox][] to other pathogen repos ([rsv][],
+[zika][], [dengue][], [hepatitisB][], [forecasts-ncov][]) primarily thru
+copying.  To [counter that
 proliferation](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1688577879947079),
 this repo was made.
 
-[monkeypox]: https://github.com/nextstrain/monkeypox
+[mpox]: https://github.com/nextstrain/mpox
 [rsv]: https://github.com/nextstrain/rsv
 [zika]: https://github.com/nextstrain/zika/pull/24
 [dengue]: https://github.com/nextstrain/dengue/pull/10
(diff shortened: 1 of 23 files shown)


ingest/vendored/README.md
victorlin, line 1: Changes to this file should be done in [nextstrain/ingest](https://github.com/nextstrain/ingest) then pulled into here.

victorlin (commented):
LGTM pending one small change which I'll address.

corneliusroemer:
If people request from `data.nextstrain.org` we should be able to do URL forwarding, shouldn't we? Wouldn't that be better than serving stale data?

Otherwise, for those requesting directly from AWS not via `data.nextstrain.org`, we have a choice between serving stale data or deleting the old file so that workflows error - each have advantages and disadvantages. Might be worth doing a code search on Github to see who uses the data and ping them.

corneliusroemer (approved):
Looks good, but I would consider setting a HTTP redirect from `https://data.nextstrain.org/files/workflows/monkeypox` to `https://data.nextstrain.org/files/workflows/mpox` if this is possible

joverlee521:
> except for a handful of locations like the #monkeypox-updates Slack notification.

Addressing this separately in #221

ingest/vendored/README.md
victorlin: Done: https://github.com/nextstrain/ingest/pull/30, b14d6fd147627ec8e34f4c84aeb987dcbb5a2838

corneliusroemer:
These are all the repos that Github could find using `https://data.nextstrain.org/files/workflows/monkeypox` that could benefit from URL redirects (17, with a handful of non-Nextstrain repos): https://github.com/search?q=data.nextstrain.org%2Ffiles%2Fworkflows%2Fmonkeypox&type=code

These are the repos that use aws s3 paths that can't be redirected (10 mostly ours): https://github.com/search?q=s3%3A%2F%2Fnextstrain-data%2Ffiles%2Fworkflows%2Fmonkeypox&type=code

joverlee521:
> If people request from data.nextstrain.org we could be able to do URL forwarding, couldn't we? 

Yes, definitely possible. Not sure if we need to do it on the Cloudfront side but it's possible to [configure S3 website redirections](https://docs.aws.amazon.com/AmazonS3/latest/userguide/how-to-page-redirect.html#advanced-conditional-redirects). Also looks like it's possible to [redirect requests for an S3 object](https://docs.aws.amazon.com/AmazonS3/latest/userguide/how-to-page-redirect.html#redirect-requests-object-metadata).

trvrb:
The [fetch and ingest workflow](https://github.com/nextstrain/mpox/actions/runs/6723041959) completed successfully. However, the artifacts ended up at s3://nextstrain-datafiles/workflows/mpox/branch/rename-to-mpox/. However, it looks good besides this. I'm assuming the best course of action is to merge this PR when it passes review and then to run fetch and ingest again from `main` so that the files are placed in the proper s3 location.

trvrb:
Thanks @victorlin for the fix with commit b14d6fd147627ec8e34f4c84aeb987dcbb5a2838. I'm going to go ahead and merge this now.

trvrb:
Fetch and ingest is running from `master` here: https://github.com/nextstrain/mpox/actions/runs/6724598765 and should update S3 files to specify s3://nextstrain-datafiles/workflows/mpox/. I'll monitor and update here when complete.

joverlee521:
We discussed redirection rules for monkeypox during today's dev chat and I added @tsibley 's [suggested redirection rules](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1698867258722699) to the `nextstrain-data` S3 bucket. 

So now access through data.nextstrain.org should redirect to mpox 
```
$ curl -I https://data.nextstrain.org/files/workflows/monkeypox/metadata.tsv.gz
HTTP/2 301 
content-length: 0
location: https://data.nextstrain.org/files/workflows/mpox/metadata.tsv.gz
date: Wed, 01 Nov 2023 23:55:24 GMT
server: AmazonS3
x-cache: Miss from cloudfront
via: 1.1 ac3f0425be668a2439884bb8cbd3ccd8.cloudfront.net (CloudFront)
x-amz-cf-pop: SFO53-C1
x-amz-cf-id: 14QBYLPsKkmgKazVcOsSE7kPxM8NRaCzXSOe-rcV6odttMGDfwU0EA==
```

Redirection for the `s3://` paths are not possible. However, based on [Cornelius's search](https://github.com/search?q=s3%3A%2F%2Fnextstrain-data%2Ffiles%2Fworkflows%2Fmonkeypox&type=code), the `s3://` paths are only used for our internal uploads and slack notifications.

trvrb:
Perfect! Thank you @joverlee521.

I can confirm that `s3://nextstrain-data/files/workflows/mpox/metadata.tsv.gz` and hence https://data.nextstrain.org/files/workflows/mpox/metadata.tsv.gz now exists as expected after running ingest above.

I deleted the `monkeypox/` directory in `files/workflows/` to reduce confusion. And as expected from your post above, https://data.nextstrain.org/files/workflows/monkeypox/metadata.tsv.gz works to retrieve the `mpox/` equivalent.

I believe the next step here is to confirm that a phylogenetics rebuild works as expected. I'll kick this off now.

trvrb:
And actually... this testing already happened with the automatic updates. Earlier today actions including https://github.com/nextstrain/mpox/actions/runs/6735424381 ran and produced updated Auspice JSONs like `mpox_all-clades.json`. I believe we're set in this case. The corresponding issue https://github.com/nextstrain/mpox/issues/168 is now closed.

PR #221, opened 2023-11-01
.github/workflows: Use centralized SLACK_CHANNELS variable

joverlee521:

Using the centralized variable makes it easier for us to update the slack channel value when needed.

As a part of this change, I've already added the SLACK_CHANNELS variable to the repo. I opted to use the channel ID instead of the channel name so that we have the option to rename the slack channel if desired.

## Related issue(s)

Related to https://github.com/nextstrain/mpox/pull/220

## Checklist

- [ ] Checks pass

### DIFF
--- a/.github/workflows/fetch-and-ingest.yaml
+++ b/.github/workflows/fetch-and-ingest.yaml
@@ -42,7 +42,7 @@ jobs:
       env: |
         NEXTSTRAIN_DOCKER_IMAGE: ${{ inputs.image }}
         GITHUB_RUN_ID: ${{ github.run_id }}
-        SLACK_CHANNELS: monkeypox-updates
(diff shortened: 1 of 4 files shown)


victorlin (approved):
I didn't realize GitHub [added support](https://github.blog/changelog/2023-01-10-github-actions-support-for-configuration-variables-in-workflows/) for non-secret variables - that's neat!

joverlee521:
> I didn't realize GitHub [added support](https://github.blog/changelog/2023-01-10-github-actions-support-for-configuration-variables-in-workflows/) for non-secret variables - that's neat!

Yup! Unlike `secrets`, the `vars` context is directly available for `jobs.<job_id>.with.<with_id>` so we can use it directly here.

PR #223, opened 2023-11-15
Fix ingest uploads

joverlee521:
Fix bug that was introduced in https://github.com/nextstrain/mpox/pull/222

The testing done on branch runs do not include the Slack notifications, so this bug was not revealed in the test run. I only noticed the workflow as failing because our internal #monkeypox-updates channel had been quiet for over a week.

This did not trigger our usual error notifications in Slack because the error occurs during the DAG building process before the start of the actual workflow run.

## Checklist

- [ ] Checks pass
- [ ] Test run in https://github.com/nextstrain/mpox/actions/runs/6882965932

### DIFF
--- a/ingest/workflow/snakemake_rules/upload.smk
+++ b/ingest/workflow/snakemake_rules/upload.smk
@@ -35,9 +35,9 @@ def _get_upload_inputs(wildcards):
     if send_notifications:
         flag_file = []
 
-        if file_to_upload == "data/genbank.ndjson":
+        if inputs["file_to_upload"] == "data/genbank.ndjson":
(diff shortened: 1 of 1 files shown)


corneliusroemer:
Thanks for spotting! Was about to investigate myself. Should we include Slack notifications to a test channel for better test coverage? Wouldn't be hard to set up a test slack channel so this type of bug is spotted more quickly next time round.

joverlee521:
> Should we include Slack notifications to a test channel for better test coverage? Wouldn't be hard to set up a test slack channel so this type of bug is spotted more quickly next time round.

Yup, see https://github.com/nextstrain/mpox/pull/224

PR #224, opened 2023-11-15
fetch-and-ingest-branch: Send Slack notications to test channel

joverlee521:
This change was motivated by the unintentional bug introduced in https://github.com/nextstrain/mpox/pull/222 that would only be triggered by using Slack notifications. This allows to test branches and send notifications to the testing channel.

As part of this change, I've added an organization level variable `TEST_SLACK_CHANNEL` that points our #scratch channel for testing Slack notifications.

## Checklist

- [x] Checks pass
- [x] Test run: https://github.com/nextstrain/mpox/actions/runs/6883650646 -> successful [Slack notifications in #scratch ](https://bedfordlab.slack.com/archives/C02LTGS6NLQ/p1700087415564329)

### DIFF
--- a/.github/workflows/fetch-and-ingest-branch.yaml
+++ b/.github/workflows/fetch-and-ingest-branch.yaml
@@ -34,6 +34,8 @@ jobs:
       runtime: aws-batch
       env: |
         NEXTSTRAIN_DOCKER_IMAGE: ${{ inputs.image }}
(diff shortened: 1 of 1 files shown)


corneliusroemer:
Nice, thanks! I had to look at docs to find where one sets that variable, it's here for reference: https://github.com/organizations/nextstrain/settings/variables/actions

image

corneliusroemer (approved):

PR #225, opened 2023-12-08
Add Nextclade dataset workflow in `/nextclade` dir

corneliusroemer:
Workflow is already being used to create v3 Nextclade datasets, merging now that it's more or less in maintenance state to reduce forking


.pre-commit-config.yaml
```diff
@@ -1,5 +1,13 @@
 exclude: '\.(tsv|fasta|gb)$|^ingest/vendored/'
 repos:
+  # Disabled for now due to bug:
+  # https://github.com/snakemake/snakefmt/issues/207
+  # - repo: https://github.com/snakemake/snakefmt
+  #   rev: v0.8.5
+  #   hooks:
+  #     - id: snakefmt
+  - repo: https://github.com/rhysd/actionlint
+    rev: v1.6.26
```
corneliusroemer, line 10:
```suggestion
    rev: v1.6.26
    hooks:
      - id: actionlint
```

PR #226, opened 2023-12-14
Version nextclade binary for transition, i.e. nextclade -> nextclade2

corneliusroemer:
Soon we will release nextclade v3. Once nextclade v3 hits bioconda, workflows that rely on v2 syntax will break. To prevent this, this PR makes it explicit that the syntax used is for v2 and that the v2 binary nextclade2 should be used.

This binary is in both the nextstrain-base docker image and in the conda-base managed conda environment.

Users who don't use either will have to add nextclade2 to the path.

Over the next few weeks, we will transition repos to use nextclade3. The reason to add nextclade2 now is so that workflows don't break if they haven't yet transitioned to nextclade3 at the time of v3 release. It also makes it easy to search which workflows need to be changed.

### DIFF
--- a/ingest/workflow/snakemake_rules/nextclade.smk
+++ b/ingest/workflow/snakemake_rules/nextclade.smk
@@ -32,7 +32,7 @@ rule align:
     threads: 4
     shell:
         """
-        nextclade run -D {input.dataset} -j {threads}   --retry-reverse-complement \
+        nextclade2 run -D {input.dataset} -j {threads}   --retry-reverse-complement \
                   --output-fasta {output.alignment}  --output-translations {params.translations} \
                   --output-insertions {output.insertions} {input.sequences}
         zip -rj {output.translations} data/translations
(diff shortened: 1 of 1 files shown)


corneliusroemer:
Merging as we will release tomorrow

joverlee521:
The ingest pipeline ran fine the last couple days but ran into an error today for the align rule:
```
[batch] [2024-01-18T08:13:06-08:00] Error: 
[batch] [2024-01-18T08:13:06-08:00]    0: ^[[91mspecified file not found in archive^[[0m
[batch] [2024-01-18T08:13:06-08:00] Location:
[batch] [2024-01-18T08:13:06-08:00]    ^[[35mpackages_rs/nextclade-cli/src/dataset/dataset_download.rs^[[0m:^[[35m65^[[0m
[batch] [2024-01-18T08:13:06-08:00] Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
[batch] [2024-01-18T08:13:06-08:00] Run with RUST_BACKTRACE=full to include source snippets.
```

It's not clear why this ran fine before, but I think this is due to the nextclade dataset command downloading the V3 dataset? 
https://github.com/nextstrain/mpox/blob/5d3a2675b5f3c2c142d8d65dd8937f093e9110be/ingest/workflow/snakemake_rules/nextclade.smk#L7

PR #247, opened 2024-05-01
Fix fstring format

joverlee521:

Follow up to https://github.com/nextstrain/mpox/pull/246

## Checklist

- [ ] Checks pass


genehack (approved):

joverlee521:
Looks like this is related to https://github.com/snakemake/snakefmt/issues/220 with using Python 3.12
We are pinned to Python 3.11 in https://github.com/nextstrain/mpox/pull/246/commits/456df4022bb114e9803e1f7dd8126be70aee3a63 so we should be okay going forward.

PR #252, opened 2024-05-09
[phylo] Reconnect traits to export

### LINKED ISSUE #251: [phylo] traits input not passed to `augur export`
In the phylogenetic workflow, the [traits input](https://github.com/nextstrain/mpox/blob/5a52c3f1a6293b2e9894efdd06dba94fc2c47c4c/phylogenetic/rules/export.smk#L73) is not passed to the `augur export` command [`--node-data` option in the shell](https://github.com/nextstrain/mpox/blob/5a52c3f1a6293b2e9894efdd06dba94fc2c47c4c/phylogenetic/rules/export.smk#L98).

 Looks like this was intentionally removed in https://github.com/nextstrain/mpox/commit/43fdaf1015a5a55b580612308c2b631bc22e9f74 to disable reconstruction of the `country` field. 

This can be confusing for users if they want traits reconstruction for their own analysis.

joverlee521:

Reconnects the `augur traits` output to `augur export` based on the presence of the `traits.columns` config parameter. 

## Related issue(s)

Resolves #251 

## Checklist

- [x] Checks pass

### DIFF
--- a/phylogenetic/rules/export.smk
+++ b/phylogenetic/rules/export.smk
@@ -70,7 +70,11 @@ rule export:
             if config.get("timetree", False)
             else "results/{build_name}/branch_lengths_no_time.json"
         ),
-        traits=build_dir + "/{build_name}/traits.json",
+        traits=(
+            build_dir + "/{build_name}/traits.json"
+            if config.get("traits", {}).get("columns", False)
+            else []
+        ),
         nt_muts=build_dir + "/{build_name}/nt_muts.json",
         aa_muts=build_dir + "/{build_name}/aa_muts.json",
         clades=build_dir + "/{build_name}/clades.json",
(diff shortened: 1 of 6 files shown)


phylogenetic/rules/export.smk
```diff
@@ -32,9 +32,9 @@ OUTPUTS:
 
 rule remove_time:
     input:
-        "results/{build_name}/branch_lengths.json",
+        build_dir + "/{build_name}/branch_lengths.json",
```
genehack, line 35: this is probably some snakemake lore i haven't absorbed yet, but i wonder: why concatenate instead of interpolate? (e.g., `"{build_dir}/{build_name}/branch_lengths.json"`)

genehack (approved):

PR #255, opened 2024-05-28
Allow addition of unreleased INRB data

jameshadfield:
These changes are to allow addition of private metadata + sequences to be spiked into the build (after filtering but prior to subsampling). They are intended for use by INRB using an excel + fasta file which they maintain. As such, these changes probably shouldn't be merged into our canonical mpox repo, but perhaps we fork the repo into the INRB organisation and maintain these (small) changes there?

If you have access to the test `xlsx` + `fasta` (see [this internal slack thread](https://bedfordlab.slack.com/archives/C03G8HQEV18/p1716871871560799)) then you can run from within the `phylogenetic` directory via:
1. `python scripts/curate_private_data.py --sequences <fasta file> --xlsx <xlsx file> --remap-columns 'accession:accession' 'accession:strain' 'collection date:date' 'province:division' 'health zone:location'   --fasta-header-idx 1`
2. `nextstrain build . --configfile defaults/clade-i/config.yaml --config private_data=true auspice_config=defaults/clade-i/auspice_config_inrb.json -f results/clade-i/good_metadata_combined.tsv auspice/mpox_clade-I.json`

**to do**
- [ ] Metadata values need updating (e.g. they have "Homo-sapiens" we have "Homo sapiens")
- [ ] Fork into INRB repo?
- [ ] Test from INRB


phylogenetic/rules/prepare_sequences.smk
```diff
@@ -74,16 +79,45 @@ rule filter:
...
+rule add_private_metadata:
+    """
+    This rule presupposes the presence of private data in the correct format
+    """
     input:
+        sequences=build_dir + "/{build_name}/good_sequences.fasta",
         metadata=build_dir + "/{build_name}/good_metadata.tsv",
+    params:
+        private_sequences="data/sequences-private.fasta",
+        private_metadata="data/metadata-private.tsv",
```
huddlej, line 100: I would make these inputs instead of params, so Snakemake can alert users immediately when they run the workflow and those files are missing.

phylogenetic/scripts/curate_private_data.py
```diff
@@ -0,0 +1,170 @@
...
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
+    parser.add_argument('--sequences', type=str, required=True, nargs='+', metavar='FASTA', help="input sequences")
+    parser.add_argument('--fasta-header-idx', type=int,
+        help='If FASTA headers are "|" separated, this index (1-based) is the accession')
+    parser.add_argument('--xlsx', type=str, required=True, help="Metadata file (Excel .xlsx format)")
+    parser.add_argument('--remap-columns', type=str, nargs='+', default=[], metavar='old:new', required=False,
+        help="Change column names. Note all column names are converted to lower case.")
+    parser.add_argument('--output-sequences', type=str, default=None, help="output sequences (default is to write to 'data/...' to then be used by our Snakemake workflows)")
+    parser.add_argument('--output-metadata', type=str, default=None, help="output metadata (default is to write to 'data/...' to then be used by our Snakemake workflows)")
```
huddlej, line 34: If you were going to keep these output arguments, you might set the default values to the relative paths to plan to use below to let the user know what the actual filenames will be before they run the script.

huddlej (commented):
This review is just of the test commands and ability of the workflow to produce a tree and not a full code review or phylogenetic review.

The test commands worked for me from an ambient Nextstrain Conda environment after I ran the following commands:

```bash
# Needed to install dependencies not in Nextstrain env.
conda install openpyxl

# Needed to make data dir in fresh clone of mpox repo.
mkdir -p data/
```

After those commands, both the test commands ran without issue and the output tree from the workflow had the private data included as expected. 🎉 

I was surprised to see that the output sequences and metadata from `curate_private_data.py` had default values instead of required values, but I see that the workflow expects those specific default names. Do you want to allow users to configure those names in the workflow config file? If they change the defaults, the workflow won't work as expected, so you could drop those output options completely and force the defaults.

Other minor comments included below.

phylogenetic/scripts/curate_private_data.py
jameshadfield: Thanks! I've dropped them here for simplicity. If we expand the remit it's easy to bring them back.

phylogenetic/rules/prepare_sequences.smk
jameshadfield: Of course! Switched to this usage. Very nice

jameshadfield:
Superseded by #287

> Do you want to allow users to configure those names in the workflow config file?

Done in #287 - thanks for the suggestion

PR #264, opened 2024-07-29
Check for mutations and clades before removing

huddlej:

Fixes a bug in the fix_tree script when trying to remove mutations or clades from lists where those values don't exist. The short-term fix is to check for the presence of the values in the lists before trying to remove them. The long term fix is to use a different tree builder than eliminates the need for a "fix tree" script.

## Checklist

- [x] Checks pass


joverlee521 (approved):
Thanks @huddlej for this fix! Changes look reasonable to me so I'm merging to resolve https://github.com/nextstrain/mpox/issues/299.

joverlee521:
Running [hmpxv1-big build](https://github.com/nextstrain/mpox/actions/runs/12892784429) with these latest changes.

Issue #272, opened 2024-08-18
Add colours for clade-Ia / Ib in all-clades build


jameshadfield:
<img width="269" alt="image" src="https://github.com/user-attachments/assets/0748f94f-0fda-4de6-b40e-ea1de4d922d1">

jameshadfield:
Will be closed by #272

jameshadfield:
Closed by #274

PR #274, opened 2024-08-23
Update clade definition and coloring

trvrb:
This PR:

1. Very slightly changes clade definition for "clade II". This results in no change to the resulting clade assignments.

Previously clade II definitions included nucleotide positions 35253 and 150970. However, neither of these positions show variation:
- [nextstrain.org/mpox/all-clades?c=gt-nuc_35253](https://nextstrain.org/mpox/all-clades?c=gt-nuc_35253)
- [nextstrain.org/mpox/all-clades?c=gt-nuc_150970](https://nextstrain.org/mpox/all-clades?c=gt-nuc_150970)

I updated these sites to 20605 and 174146 which show variation between clades I and II.

2. The color ordering lacked entries for clade Ia and clade Ib. This resulted in gray colors for these two clades at nextstrain.org/mpox/all-clades.

The resulting `all-clades` build looks like:

Screenshot 2024-08-23 at 2 20 28 PM

I don't think this should be too controversial, but could either @jameshadfield or @corneliusroemer take a very quick look?

### DIFF
--- a/phylogenetic/defaults/clades.tsv
+++ b/phylogenetic/defaults/clades.tsv
@@ -11,9 +11,9 @@ clade Ia	nuc	35352	A
 clade Ib	clade	clade I
 clade Ib	nuc	19455	T
 
+clade II	nuc	20605	A
 clade II	nuc	86502	G
-clade II	nuc	150970	A
-clade II	nuc	35352	C
+clade II	nuc	174146	C
 
 clade IIa	clade	clade II
 clade IIa	nuc	54013	G
--- a/phylogenetic/defaults/color_ordering.tsv
+++ b/phylogenetic/defaults/color_ordering.tsv
@@ -16884,7 +16884,8 @@ recency	New
 
 ################
 
-
+clade_membership	Ia
+clade_membership	Ib
 clade_membership	I
 clade_membership	II
 clade_membership	IIa


jameshadfield (approved):
I can't speak to the previous clade-II mutations, but the new ones are unique to clade-II in a just-built mpxv tree. Note that there are no tips which are labelled as clade-II or clade-I (only internal nodes) which has a few oddities in auspice (see https://github.com/nextstrain/auspice/issues/1275) - they appear as distinct colours but not in the legend, you can't filter to them directly etc. 

"The color ordering lacked entries for clade Ia and clade Ib" (in the mpxv build) is #272  which is fixed here (and the clade-I build also gets non-default colours as a result).

corneliusroemer (approved):
Thanks for double checking the clades TSV.

@trvrb I think you might have made a digit swap when hand-typing 35352/35253 - it's 35352 in the clades.tsv and your PR mentions 35253 (an anagram with the 3rd and 5th digit swapped).

What used to be in the clades.tsv has diversity at that site:

Google Chrome 2024-08-27 07 57 16
https://nextstrain.org/mpox/all-clades?c=gt-nuc_35352&gmax=36253&gmin=34253

But given it's only in Ia not I it's best to remove/replace in any case.

150970 might have shifted to 150968 (which does show I/II diversity A/T due to changes in Nextclade alignment - good spot.

PR #278, opened 2024-09-27
Clade-I tweaks

jameshadfield:
Clade-I specific changes to how geographic metadata is passed through and how clades are assigned


jameshadfield:
I'd like to merge this ASAP and will do so tomorrow if there are no objections. The only changes are to clade-I builds and they fix the currently buggy clade labeling as well as adding metadata to reflect the richness of the newly submitted INRB data.

PR #279, opened 2024-09-27
Update geo-fixes

jameshadfield:
See added comments in TSV

- [ ] Checks pass

### DIFF
--- a/ingest/defaults/geolocation-rules.tsv
+++ b/ingest/defaults/geolocation-rules.tsv
@@ -14,3 +14,9 @@ Europe/Italy/Fvg/Gorizia	Europe/Italy/Friuli Venezia Giulia/Gorizia
 # Unclear which location is the real location
 Europe/Netherlands/Utrecht/Rotterdam	Europe/Netherlands//
 North America/USA/Washington/Dc	North America/USA/Washington DC/
+# Following renamings as part of work on INRB Clade-I builds
+Africa/Democratic Republic of the Congo/*/Nyagezi	Africa/Democratic Republic of the Congo/*/Nyangezi
+# Note that in our master geolocation rules we replace "Sud-Kivu" with "Sud Kivu" (similarly for "Nord-Kivu")
+# however the INRB tends to use "Sud-Kivu". We can't add a rule here to change it back (`augur curate` doesn't
+# allow it) so we enforce the use of "Sud Kivu" in INRB data as well.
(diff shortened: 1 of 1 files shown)


jameshadfield:
I'm going to merge this ASAP (post-merge review welcome) as none of our mpox builds currently have sub-country level metadata included. #278 will add this to the Clade-I build, and we are using it for INRB builds as well.

ingest/defaults/geolocation-rules.tsv
```diff
@@ -14,3 +14,9 @@ Europe/Italy/Fvg/Gorizia	Europe/Italy/Friuli Venezia Giulia/Gorizia
 # Unclear which location is the real location
 Europe/Netherlands/Utrecht/Rotterdam	Europe/Netherlands//
 North America/USA/Washington/Dc	North America/USA/Washington DC/
+# Following renamings as part of work on INRB Clade-I builds
+Africa/Democratic Republic of the Congo/*/Nyagezi	Africa/Democratic Republic of the Congo/*/Nyangezi
+# Note that in our master geolocation rules we replace "Sud-Kivu" with "Sud Kivu" (similarly for "Nord-Kivu")
+# however the INRB tends to use "Sud-Kivu". We can't add a rule here to change it back (`augur curate` doesn't
+# allow it) so we enforce the use of "Sud Kivu" in INRB data as well.
```
joverlee521, line 21: I was curious what you meant by "augur curate doesn't allow it"...Because we just concatenate these to the "general" geolocation rules, the conflicting rules raises the [CyclicGeolocationRulesError](https://github.com/nextstrain/augur/blob/master/augur/curate/apply_geolocation_rules.py#L167-L170) in `augur curate apply-geolocation-rules`. 

Not sure if this is something we should handle _within_ Augur or if the workflow should be improved to not blindly concatenate the geolocation rules...

jameshadfield: > Not sure if this is something we should handle within Augur...

If we're going to continue with the pattern of a master-list plus some repo-specific additions it'd be nice if those additions could function as overrides. I.e. if the master list replaces X with Y, we could add an override of Y to X (or Y to Z etc) in a repo-specific manner.

joverlee521: Yeah, that makes sense! Tracking in https://github.com/nextstrain/augur/issues/1648.

PR #284, opened 2024-10-31
Fix Clade I build

joverlee521:

Use `--query` to filter for clade I build since we have migrated to Nextclade v3 that includes clades I, Ia, and Ib.

## Related issue(s)

Follow up to https://github.com/nextstrain/mpox/pull/281
Resolves https://github.com/nextstrain/mpox/issues/283

## Checklist

- [x] Checks pass
- [x] [Trial Clade I build](https://github.com/nextstrain/mpox/actions/runs/11616384964)

### DIFF
--- a/phylogenetic/defaults/clade-i/config.yaml
+++ b/phylogenetic/defaults/clade-i/config.yaml
@@ -19,14 +19,14 @@ auspice_name: "mpox_clade-I"
 filter:
   min_date: 1900
   min_length: 100000
-  exclude_where: 'clade!=I'
 
 
-### We don't want to subsample, so specify a config which is essentially a no-op
+### Filter to only Clade I sequences
 subsample:
   everything:
     group_by: ""
     sequences_per_group: ""
+    other_filters: "--query \"clade in ['I', 'Ia', 'Ib']\""
 
 ## align
 max_indel: 10000
--- a/phylogenetic/defaults/clade-i/config.yaml
+++ b/phylogenetic/defaults/clade-i/config.yaml
@@ -26,7 +26,7 @@ subsample:
   everything:
     group_by: ""
     sequences_per_group: ""
-    other_filters: "--query \"clade in ['I', 'Ia', 'Ib']\""
+    query: "'clade in [\"I\", \"Ia\", \"Ib\"]'"
 
 ## align
 max_indel: 10000
--- a/phylogenetic/rules/prepare_sequences.smk
+++ b/phylogenetic/rules/prepare_sequences.smk
@@ -98,6 +98,11 @@ rule subsample:
         sequences_per_group=lambda w: config["subsample"][w.sample][
             "sequences_per_group"
         ],
+        query=lambda w: (
+            f"--query {config['subsample'][w.sample]['query']}"
+            if "query" in config["subsample"][w.sample]
+            else ""
+        ),
         other_filters=lambda w: config["subsample"][w.sample].get("other_filters", ""),
         exclude=lambda w: (
             f"--exclude-where {' '.join([f'lineage={l}' for l in config['subsample'][w.sample]['exclude_lineages']])}"
@@ -113,6 +118,7 @@ rule subsample:
             --output-strains {output.strains} \
             {params.group_by} \
             {params.sequences_per_group} \
+            {params.query} \
             {params.exclude} \
             {params.other_filters} \
             --output-log {output.log}


phylogenetic/defaults/clade-i/config.yaml
```diff
@@ -19,14 +19,14 @@ auspice_name: "mpox_clade-I"
...
-  exclude_where: 'clade!=I'
 
 
-### We don't want to subsample, so specify a config which is essentially a no-op
+### Filter to only Clade I sequences
 subsample:
   everything:
     group_by: ""
     sequences_per_group: ""
+    query: "'clade in [\"I\", \"Ia\", \"Ib\"]'"
```
victorlin, line 29: _(non-blocking)_

Suggestion: use set instead of list because `item in Set` is O(1) while `item in List` is O(n). The different in speed might be negligible for the Clade I build so I wouldn't consider this blocking, but just wanted to flag a potential improvement that can be made elsewhere too.

```suggestion
    query: "'clade in {\"I\", \"Ia\", \"Ib\"}'"
```

Example:

```
In [1]: %timeit "II" in {"I", "Ia", "Ib"}
12.7 ns ± 0.0867 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)

In [2]: %timeit "II" in ["I", "Ia", "Ib"]
38.1 ns ± 0.52 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

Note: creating a set is more costly than creating a list with the same items, but even in the worst case example above where it is created on every loop (unless there are per-run optimizations), the set method is still faster. For `augur filter`, the set should only be created once per metadata chunk so it should be less of a concern.

joverlee521:
Trial build deployed to https://nextstrain.org/staging/trial/fix-clade-i/mpox/clade-I

phylogenetic/defaults/clade-i/config.yaml
```diff
@@ -19,14 +19,14 @@ auspice_name: "mpox_clade-I"
 filter:
   min_date: 1900
   min_length: 100000
-  exclude_where: 'clade!=I'
```
victorlin, line 22: Sorry, I'm still confused...

After reading https://github.com/nextstrain/mpox/issues/283#issuecomment-2448388791, I was under the impression that the change would be to replace `subsample.<sample>.exclude_where: 'clade!=I'` with `subsample.<sample>.query: clade in I, Ia, Ib`. But now I see that it's replacing `filter.exclude_where: 'clade!=I'` with `subsample.<sample>.query: clade in I, Ia, Ib`. Why not `filter.query` to keep consistent with having all filters in `filter`?

phylogenetic/defaults/clade-i/config.yaml
joverlee521: Hmm, when I run the build with this change, I get an error from Pandas

```
ERROR: Internal Pandas error when applying query:
	'Set' nodes are not implemented
Ensure the syntax is valid per <https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#indexing-query>.
```

victorlin: I should've checked with pandas.query first, that's a bummer...

phylogenetic/defaults/clade-i/config.yaml
joverlee521: Good question! 

This is my interpretation of the workflow based on the other mpox build configs (e.g. [hmpxv_big](https://github.com/nextstrain/mpox/blob/4ce0b5bdc6a0c21d1eaee95ff7d17725b1470b27/phylogenetic/defaults/hmpxv1_big/config.yaml#L19-L36)):

1. Use `filter` to filter for "good" sequences (by date, length, QC)
2. Use `subsample` to filter and subsample by clade and lineages

The clade I build actually deviated from this pattern by [adding `filter.exclude_where`](https://github.com/nextstrain/mpox/commit/1140955fae354b87b3c92e2d5a8191898a6f1312#diff-041205ab957fdd3aac682dffdc4b564dba6bbfd077747724676d6a60f46887ad). Using `subsample.<sample>.query` here to bring it back to the pattern used by other builds.

victorlin (approved):

joverlee521:
I compared the trial build to the current live build and I'm confident in these changes so I'm going to merge. 

---

Summary of comparison:

The trial build has one extra sequence (PQ221596/23MPX0047C), but all other sequences are the same based on accession. PQ221596 was previously unassigned with Nextclade v2, but is now assigned to clade Ia with Nextclade v3. So inclusion of PQ221596 in the trial build is expected.

I noticed that the trial build clade assignments are slightly different than the live build in [tangletree](https://nextstrain.org/mpox/clade-I:staging/trial/fix-clade-i/mpox/clade-I?c=clade_membership):

Screenshot 2024-10-31 at 11 37 58 AM

I re-ran the build locally and the clade assignments are the same as the live build:

Screenshot 2024-10-31 at 12 20 16 PM

I'm going to attribute this to some stochastic tree building weirdness that is _**not**_ a result of any changes in this PR.

PR #288, opened 2024-11-14
Disable transmission lines again, they are too often wrong

corneliusroemer:


genehack (approved):

corneliusroemer:
We generally don't enable these transmission lines on Nextstrain builds, hence disabling for now again

PR #289, opened 2024-11-14
Add new clades to phylogenetic

corneliusroemer:
Follows https://github.com/nextstrain/mpox/pull/241 as template


joverlee521 (approved):
Not really familiar with mpox lineages, but changes look reasonable to me. 

I do wonder if it'd be worth it to consolidate the references files such as clade definitions and color ordering between phylogenetic and nextclade workflows.

corneliusroemer:
Maybe worth, but I'm not going to do it now if I only edit those files twice a year 😄

corneliusroemer:
I just designated those lineages over at the designation repo so this is just the logical followup that I always have to do: https://github.com/mpxv-lineages/lineage-designation/pull/45

PR #312, opened 2025-04-30
phylogenetic: Add author coloring & filter for all builds

joverlee521:

Since <https://github.com/nextstrain/mpox/pull/295>, the metadata includes the abbreviated authors list in the "authors" column of the metadata.tsv. However, `augur export` renames `authors` -> `author`,¹ so we must use `author` in the auspice_config.json to include it as coloring and filters.

Resolves <https://github.com/nextstrain/mpox/issues/311>

¹ <https://github.com/nextstrain/augur/blob/be4215b17c03796bd8c75fb5fc3d6b5bb4c83438/augur/export_v2.py#L172>

## Checklist

- [x] Checks pass

### DIFF
--- a/phylogenetic/defaults/hmpxv1_big/auspice_config.json
+++ b/phylogenetic/defaults/hmpxv1_big/auspice_config.json
@@ -60,6 +60,11 @@
       "key": "date",
       "title": "Collection date",
       "type": "categorical"
+    },
+    {
+      "key": "author",
+      "title": "Authors",
+      "type": "categorical"
     }
   ],
   "geo_resolutions": [
@@ -77,6 +82,7 @@
     "country",
     "region",
     "recency",
-    "host"
+    "host",
+    "author"
   ]
 }
--- a/phylogenetic/defaults/clade-i/auspice_config.json
+++ b/phylogenetic/defaults/clade-i/auspice_config.json
@@ -92,8 +92,8 @@
       "type": "categorical"
     },
     {
-      "key": "abbr_authors",
-      "title": "Abbreviated Authors",
+      "key": "author",
+      "title": "Authors",
       "type": "categorical"
     },
     {
@@ -118,6 +118,7 @@
     "country",
     "region",
     "recency",
-    "host"
+    "host",
+    "author"
   ]
 }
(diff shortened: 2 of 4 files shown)


jameshadfield:
> However, augur export renames authors -> author,¹ so we must use author in the auspice_config.json to include it as coloring and filters.

Our [config parsing code](https://github.com/nextstrain/augur/blob/be4215b17c03796bd8c75fb5fc3d6b5bb4c83438/augur/util_support/auspice_config.py#L134-L145) should also change "authors" to "author" (keeping track of all these changes is painful!), but in this situation we were using "abbr_authors" (for clade-i) and didn't have "author(s)" in any of the other configs.

phylogenetic/defaults/hmpxv1_big/auspice_config.json
```diff
@@ -77,6 +82,7 @@
     "country",
     "region",
     "recency",
-    "host"
+    "host",
+    "author"
```
joverlee521, line 86: Continuing from @jameshadfield 's [commment](https://github.com/nextstrain/mpox/pull/312#issuecomment-2844098401)

> > However, augur export renames authors -> author,¹ so we must use author in the auspice_config.json to include it as coloring and filters.
> 
> Our [config parsing code](https://github.com/nextstrain/augur/blob/be4215b17c03796bd8c75fb5fc3d6b5bb4c83438/augur/util_support/auspice_config.py#L134-L145) should also change "authors" to "author" (keeping track of all these changes is painful!), but in this situation we were using "abbr_authors" (for clade-i) and didn't have "author(s)" in any of the other configs.

Thanks for pointing to the auspice config parsing! I had initially added "authors" for both colorings and filters, but saw this warning from export about the filter 

> WARNING:  The filter "authors" does not appear as a property on any tree nodes.

So I figured it'd be better to keep both as "author" to be consistent. Sounds like this should be a treated as a bug in the auspice config parsing, i.e. it should also convert "authors" -> "author" for filters as well.

jameshadfield: > Sounds like this should be a treated as a bug in the auspice config parsing, i.e. it should also convert "authors" -> "author" for filters as well.

Yes - I think that's right. (The config parsing code is new, but I'm almost certain the bug was pre-existing.) When I said 

> keeping track of all these changes is painful!

I meant for both users and developers!

joverlee521: Tracking the bug in https://github.com/nextstrain/augur/issues/1803. 

I'll merge this PR to add the coloring/filter for the mpox builds and we can update the configs after the bug fix if desired.

PR #321, opened 2025-06-25
phylogenetic workflow updates

joverlee521:

Various phylogenetic workflow updates ahead of adding support for `nextstrain run`. The main changes are making things more configurable via config files. See commits for details.

I think the most controversial change will be https://github.com/nextstrain/mpox/commit/f1ed6e987adea66e275dabf4ed9c1278a6561065, which switches to the generic subsample config schema. Seeking feedback on whether this is the correct direction.

## Related issue(s)

Resolves https://github.com/nextstrain/mpox/issues/273

## Checklist

- [x] Checks pass
- [x] Update CHANGELOG


phylogenetic/defaults/clade-i/config.yaml
```diff
@@ -2,8 +2,11 @@ reference: "defaults/clade-i/reference.fasta"
 genome_annotation: "defaults/clade-i/genome_annotation.gff3"
 genbank_reference: "defaults/clade-i/reference.gb"
 include: "defaults/clade-i/include.txt"
+exclude: "defaults/exclude_accessions.txt"
 clades: "defaults/clades.tsv"
 lat_longs: "defaults/lat_longs.tsv"
+color_ordering: "defaults/color_ordering.tsv"
```
jameshadfield, line 8: Discussion question prompted by the similarity between these configs.

Did we consider unifying the currently independent workflow invocations (clade-i, hmpv1 etc) into a single workflow parameterised by wildcards? Or has that ship sailed long ago? An alternate direction would be config overlays.

phylogenetic/defaults/hmpxv1/config.yaml
```diff
@@ -24,51 +24,52 @@ filter:
...
-  b1:
-    group_by: "--group-by country year"
-    sequences_per_group: "--subsample-max-sequences 300"
-    other_filters: "--exclude-where outbreak!=hMPXV-1 clade!=IIb"
+  non_b1: >-
+    --group-by lineage year country
+    --sequences-per-group 50
+    --exclude-where
+    outbreak!=hMPXV-1
+    clade!=IIb
```
jameshadfield, line 35: [minor] can you increase the indentation of the values following `--exclude-where`, or does the mixed indentation not work with YAML folded blocks?

phylogenetic/defaults/clade-i/config.yaml
```diff
@@ -23,10 +23,8 @@ filter:
 
 ### Filter to only Clade I sequences
 subsample:
-  everything:
-    group_by: ""
-    sequences_per_group: ""
-    query: "'clade in [\"I\", \"Ia\", \"Ib\"]'"
+  everything: >-
+    --query 'clade in ["I", "Ia", "Ib"]'
```
jameshadfield, line 30: > I think the most controversial change will be https://github.com/nextstrain/mpox/commit/f1ed6e987adea66e275dabf4ed9c1278a6561065, which switches to the generic subsample config schema. Seeking feedback on whether this is the correct direction.

I think the changes here are good - as you note, they align with [the linked docs](https://docs.nextstrain.org/en/latest/guides/bioinformatics/filtering-and-subsampling.html#generalizing-subsampling-in-a-workflow).  

These changes are backwards incompatible, so I'd suggest adding a check to the `config.smk` (see other comment thread) which enforces that `subsample.<sample>` must be a string value (and points to a piece of documentation explaining what to change if it's not).

jameshadfield (changes requested):
Changes look generally good - left a number of questions and discussion points in threads.

There are no updates to the [build-configs/inrb workflow](https://github.com/nextstrain/mpox/tree/master/phylogenetic/build-configs/inrb). We should endeavour to keep this up to date. I have some testing files if that would help?

phylogenetic/defaults/clade-i/config.yaml
joverlee521: > An alternate direction would be config overlays.

We tried [config overlays in the past](https://github.com/nextstrain/mpox/pull/173), but that got reverted in https://github.com/nextstrain/mpox/commit/50194209c1163d0d3e1f7dbe22a228fd7ce27258. So I think that's a no-go.

> Did we consider unifying the currently independent workflow invocations (clade-i, hmpv1 etc) into a single workflow parameterised by wildcards? Or has that ship sailed long ago? 

I like the simplicity of the single build config, but we are moving towards most pathogens having multi-build wildcard configs. So maybe the ship has not sailed if we think it's worth reworking this config. 

With `nextstrain run` in mind, there would then only be a single `phylogenetic` workflow and users would have to know the single target for the single clade build OR provide a config overlay that can specify a single clade build.

phylogenetic/defaults/hmpxv1/config.yaml
joverlee521: In YAML folded blocks, any extra indentation adds a newline.

phylogenetic/defaults/clade-i/config.yaml
joverlee521: Sounds good. I'll consolidate all the config munging and add some config validation.

phylogenetic/defaults/hmpxv1/config.yaml
jameshadfield: Bummer!

phylogenetic/defaults/clade-i/config.yaml
jameshadfield: It'd be possible to make these changes backwards compatible by adding a function to convert the old subsampling config syntax to the new one -- the code's already there to do this, it just needs a little rearranging. I'll leave it up to you to decide how important it is keep things backwards compatible for this repo.

phylogenetic/rules/annotate_phylogeny.smk
```diff
@@ -90,7 +90,7 @@ rule traits:
     output:
         node_data=build_dir + "/{build_name}/traits.json",
     params:
-        columns=config["traits"]["columns"],
+        columns=as_list(config["traits"]["columns"]),
```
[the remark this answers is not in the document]
joverlee521: I decided to keep things backwards compatible and centralized everything config.smk in https://github.com/nextstrain/mpox/pull/321/commits/06af354c107c593c4e7027f5ab57f04cb6246635

phylogenetic/rules/construct_phylogeny.smk
```diff
@@ -40,6 +40,12 @@ rule tree:
             --nthreads {threads}
         """
 
+ROOT_FLAG = "--root "
+if config.get("treefix_root", "").startswith(ROOT_FLAG):
+    print(f"WARNING: config['treefix_root'] no longer requires the flag {ROOT_FLAG!r}; "
+          f"removing from flag from the config param.", file=sys.stderr)
+    config["treefix_root"] = config["treefix_root"][len(ROOT_FLAG):]
+
```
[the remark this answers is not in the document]
joverlee521: Centralized in config.smk in https://github.com/nextstrain/mpox/pull/321/commits/06af354c107c593c4e7027f5ab57f04cb6246635

phylogenetic/defaults/clade-i/config.yaml
joverlee521: I decided to keep things backwards compatible and centralized everything config.smk in https://github.com/nextstrain/mpox/pull/321/commits/06af354c107c593c4e7027f5ab57f04cb6246635

phylogenetic/defaults/hmpxv1/config.yaml
```diff
@@ -24,51 +24,52 @@ filter:
...
-      - F.5
-      - F.6
-      - B.1.21
-      - B.1.22
-      - B.1.23
-  b1:
-    group_by: "--group-by country year"
-    sequences_per_group: "--subsample-max-sequences 300"
-    other_filters: "--exclude-where outbreak!=hMPXV-1 clade!=IIb"
+  non_b1: >-
```
joverlee521, line 30: Ah, with adding support for `nextstrain run`, I'm remembering that this [subsample schema does not work well with config overlays](https://docs.google.com/document/d/1YQ2DlCmecp0BxNYu0C0SI641aoz5JItTVhSRWrRO3HU/edit?tab=t.0#heading=h.1psg9viuaio9) because Snakemake merges dictionaries. 

Will have to think on how to make this both overridable and extendable...

joverlee521: One option would be similar to the proposed multi-input config params in https://github.com/nextstrain/zika/pull/80.
The `subsample` param will be changed to a list and there will be a new `additional_subsample` param:

```yaml
subsample:
  - name: non_b1
    args: >-
      --group-by lineage year country
      --sequences-per-group 50
      ...
  - name: b1
    args: >-
      --group-by country year
      --subsample-max-sequences 300
      ...

additional_subsample: []
```

Then config overlays can either completely override the subsample groups by using the `subsample` config param or define additional subsample groups with `additional_subsample`.

phylogenetic/defaults/clade-i/config.yaml
jameshadfield: Haven't tested the code, but the approach looks great and reads nicely.

phylogenetic/defaults/hmpxv1/config.yaml
joverlee521: > For mpox this approach seems nice. My instinct tells me it won't work for a multi-build config (e.g. [ncov!](https://github.com/nextstrain/ncov/blob/41cf6470d3140963ff3e02c29241f80ae8ed9c33/nextstrain_profiles/nextstrain-gisaid/builds.yaml#L154-L326)). Do you think it's worth solving both now, or running with this approach for the 1-config-1-build style repos and leaving the multi-build-configs until later?

Yeah, this could get hairy for multi-build configs...I do think it'd be worth trying to find a solution that works for both. Or at least an approach that doesn't make converting a single build to a multi-build config too difficult.

jameshadfield: > The config parsing in the workflow will just have to know to drop the null subsample groups.

A way to nullify config values has been something many of us have been pushing for, so no objections there.

For me the salient question is around the user experience for designing config overlays. If you want to go the (dict + nullification) direction, which is programmatically simpler, then a user has to know the keys in the (default) `subsample` dict, and we have to have a story for communicating this without saying "read the src". (Aside: this is exactly what I did for the inputs interface - I implemented it as a single dict with `∅` used for nullification - but for the reason I just described I went with the `{additional,}inputs` design.) 

What to do right now for mpox? I'd go for a single dict + nullification. All that's needed to add is the nullification aspect. At the current time you have to read the src to build a config overlay, so that's not a problem at the moment.

joverlee521: Thanks for the feedback @jameshadfield! I'll go with dict + nullification here.

Surfacing the default `subsample` dict for the user the nullify could be part of the larger [issue for logging config values](https://github.com/nextstrain/pathogen-repo-guide/issues/18)

joverlee521:
Pushed up changes to move filter query into config param in [1021e9c](https://github.com/nextstrain/mpox/pull/321/commits/1021e9c44bf3e3e4c6076ca7190067d6ba55d37d) (which was the original purpose of this PR but somehow I forgot about it in all the other changes 🤦‍♀️ )

If there are no other comments, I'll merge this on Monday.

.pre-commit-config.yaml
```diff
@@ -1,4 +1,4 @@
-exclude: '\.(tsv|fasta|gb)$|^ingest/vendored/'
+exclude: '\.(tsv|fasta|gb)$|^ingest/vendored/|^phylogenetic/rules/config.smk'
```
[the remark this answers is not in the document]
joverlee521: I was not a fan of the auto-reformatting done in https://github.com/nextstrain/mpox/commit/6462c34344075da275641435b71af4c3b5886460 and I didn't want to dig into how to force `snakefmt` to _not_ edit those files.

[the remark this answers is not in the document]
genehack: Totally fair! 😆

Issue #314, opened 2025-06-11
Support for `nextstrain run`


joverlee521:
See required workflow changes in https://github.com/nextstrain/zika/pull/85.

## Considerations

The one big difference between mpox build and the zika build is we define multiple builds via different config files in `phylogenetic/defaults`. This doesn't really mesh well with `nextstrain run` since the passed config files are expected to be in the user's analysis directory and _not_ the source config files. This was previously discussed on [Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1732568407123369) and [GitHub](https://github.com/nextstrain/measles/pull/55#discussion_r1878865458) where the big takeaway is to define named workflows with a separate `Snakefile` as an entrypoint for each workflow.

This would look something like this:
```
phylogenetic/
├── clade-i
│   ├── Snakefile
│   └── config.yaml
├── clade-iib (currently called hmpxv1)
│   ├── Snakefile
│   └── config.yaml
├── lineage-b.1 (currently called hmpxv1_big)
│   ├── Snakefile
│   └── config.yaml
├── mpxv
│   ├── Snakefile
│   └── config.yaml
├── rules
│   ├── annotate_phylogeny.smk
│   ├── construct_phylogeny.smk
│   ├── export.smk
│   ├── main.smk
│   └── prepare_sequences.smk
└── ... (other files)
```
The `phylogenetic/Snakefile` file gets moved to `rules/main.smk` and each `phylogenetic/<workflow>/Snakefile` would import the `main.smk`, similar to [avian-flu's Snakefile](https://github.com/nextstrain/avian-flu/blob/master/genome-focused/Snakefile).  

Then to run the hmpxv1/clade-iib workflow 

_with `nextstrain build`_

    nextstrain build phylogenetic --snakefile clade-iib/Snakefile

_with `nextstrain run`_

    nextstrain run mpox phylogenetic/clade-iib <analysis-dir>

### Config files 

There are shared config files used for all builds (e.g. description.md) and there are also build specific files (e.g. auspice_config.json). I think these can continue to live within `phylogenetic/defaults/` but we'll see how the `resolve_config_path` works with this structure.

PR #327, opened 2025-07-09
(WIP) phylogenetic: Support `nextstrain run`

joverlee521:

Workflow updates to support `nextstrain run`. Since there are multiple builds,  I've created a separate entrypoint/Snakefile per build.

This allows users to use `nextstrain run` without know the internal config file  locations:
```
nextstrain run mpox phylogenetic/clade-i <analysis-dir>
```

For users who still want to use `nextstrain build`, the command is now 
```
nextstrain build phylogenetic -s clade-i/Snakefile
```

## Discussion

Currently a draft PR mainly for discussion of the changes needed to support `nextstrain run`.

I want to mainly gather thoughts on the 4 extra directories/Snakefiles as separate workflow entrypoints. I _think_ the long term goal is for `nextstrain run` to be smarter about picking workflows to run, e.g. imagine defining the workflows within `nextstrain-pathogen.yaml`. 

## Related issue(s)

Part of #314 

## Checklist

- [ ] Checks pass
- [ ] Update changelog

### DIFF
--- a/phylogenetic/clade-iib/Snakefile
+++ b/phylogenetic/clade-iib/Snakefile
@@ -1,0 +1,10 @@
+configfile: os.path.join(workflow.basedir, "../defaults/hmpxv1/config.yaml")
+
+if os.path.exists("config.yaml"):
+    configfile: "config.yaml"
+
+
+include:"../rules/main.smk"
+
+rule _all:
+    input: rules.all.input
--- a/phylogenetic/rules/config.smk
+++ b/phylogenetic/rules/config.smk
@@ -7,6 +7,23 @@ from textwrap import dedent, indent
 from typing import Union
 
 
+include: "../../shared/vendored/snakemake/config.smk"
+
+
+def phylo_resolve_config_path(path: str) -> Callable[[Wildcards], str]:
+    """
+    Wrapper around the shared `resolve_config_path` to force the default directory
+    to be `phylogenetic/defaults`. This is necessary because the entry point for
+    each build is nested within phylogenetic (e.g. phylogenetic/clade-i/Snakefile).
+    """
+    PHYLO_DEFAULTS_DIR = os.path.normpath(os.path.join(workflow.current_basedir, "../defaults"))
+    # Strip the `defaults/` prefix to be backwards compatible with older configs
+    # This is necessary in this wrapper because we are providing a custom defaults dir
+    # which skips the handling of the defaults/ prefix within resolve_config_path.
+    path = path.removeprefix("defaults/")
+    return resolve_config_path(path, PHYLO_DEFAULTS_DIR)
+
+
 def as_list(config_param: Union[list,str]) -> list:
     if isinstance(config_param, list):
         return config_param
--- a/phylogenetic/rules/main.smk
+++ b/phylogenetic/rules/main.smk
@@ -10,10 +10,6 @@ if version.parse(augur_version) < version.parse(min_augur_version):
     )
     sys.exit(1)
 
-if not config:
-
-    configfile: "defaults/hmpxv1/config.yaml"
-
 
 build_dir = "results"
 auspice_dir = "auspice"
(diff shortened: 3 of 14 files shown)


phylogenetic/clade-iib/Snakefile
jameshadfield, line 1: If we're going to rename hmpxv1 to clade-iib (which seems fine 🤷 I don't have much context here) then we should go the whole way and rename everything - `defaults/hmpxv1` etc. (Perhaps you plan to do this -- it's a draft PR after all -- if so this can just serve as a reminder)

phylogenetic/clade-iib/Snakefile
```diff
@@ -0,0 +1,14 @@
+configfile: os.path.join(workflow.basedir, "../defaults/hmpxv1/config.yaml")
```
jameshadfield, line 1: I think it's cleaner and easier to understand if we [co-locate the `config.yaml` with the Snakefile](https://github.com/nextstrain/avian-flu/tree/master/genome-focused), i.e.

```console
$ tree phylogenetic/clade-iib
clade-iib
├── Snakefile
└── config.yaml
```

This raises the question of where to store clade-iib specific config files: `phylogenetic/defaults/clade-iib` or `phylogenetic/clade-iib/defaults` (or some variant thereof). If we look at some values in the config YAML (swapping hmpxv1 for clade-iib for clarity) we have things like:

```yaml
color_scheme: "color_schemes.tsv"
auspice_config: "clade-iib/auspice_config.json"
```

which isn't very ergonomic from an external users point of view. I think something like the following is cleaner:

```yaml
color_scheme: "color_schemes.tsv"
auspice_config: "auspice_config.json"
```

```console
$ tree
phylogenetic
├── defaults
│   └── color_schemes.tsv
├── clade-iib
│   ├── Snakefile
│   ├── config.yaml
│   ├── defaults            # maybe use a defaults subdir?
│   │   └── auspice_config.json
│   └── auspice_config.json # or maybe just do this?
```

Where you have a path resolution of (1) analysis dir, (2) clade-iib, (3) phylogenetic.

phylogenetic/rules/config.smk
```diff
@@ -7,6 +7,23 @@ from textwrap import dedent, indent
...
+    each build is nested within phylogenetic (e.g. phylogenetic/clade-i/Snakefile).
+    """
+    PHYLO_DEFAULTS_DIR = os.path.normpath(os.path.join(workflow.current_basedir, "../defaults"))
+    # Strip the `defaults/` prefix to be backwards compatible with older configs
+    # This is necessary in this wrapper because we are providing a custom defaults dir
+    # which skips the handling of the defaults/ prefix within resolve_config_path.
+    path = path.removeprefix("defaults/")
+    return resolve_config_path(path, PHYLO_DEFAULTS_DIR)
+
+
```
jameshadfield, line 26: I still wish we got rid of all this magic around `defaults/` and the path in the `config.yaml` was the path used, relative to an ordered set of "base directories" which are easily documented. I understand people don't want external users to have to make a "defaults" directory, but if we used "config" instead I think there'd be less objection. 

Secondly I still think `phylo_resolve_config_path` is overly verbose and leaves the door open for unintentionally using `resolve_config_path` (which is still in the namespace). I continue to think the following is preferable:

```py
include: "../../shared/vendored/snakemake/config.smk"

_shared_resolve_config_path = resolve_config_path

def resolve_config_path(path: str) -> Callable[[Wildcards], str]:
    return _shared_resolve_config_path(...)
```

jameshadfield (commented):
> I want to mainly gather thoughts on the 4 extra directories/Snakefiles as separate workflow entrypoints. I think the long term goal is for nextstrain run to be smarter about picking workflows to run, e.g. imagine defining the workflows within nextstrain-pathogen.yaml.

Regarding the 4 extra Snakefiles I have two thoughts. Firstly, it's great that each repo has the power to do its own thing, and that it can expose ~any Snakefile (with the necessary path wrangling etc) as an interface to our runner and leverage the power that brings (cloud compute, runtimes etc). Secondly, I imagine you want to decide on a pattern that people can just copy-paste for multi-build workflows and it seems like this approach is in direct competition with using wildcards to parameterise builds. (The long-term plans for `nextstrain-pathogen.yaml` are opaque to me.)

phylogenetic/rules/config.smk
joverlee521: Yeah, this is only necessary because of the nested Snakefiles. If we keep the main `phylogenetic/Snakefile` then we could just directly use the shared `resolve_config_path`.

phylogenetic/rules/main.smk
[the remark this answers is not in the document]
jameshadfield: Yeah, this is why I used [pathogen-repo-ci.yaml@v0](https://github.com/nextstrain/avian-flu/blob/7123c7560501dc3d63fd4edf91461a02dda67d37/.github/workflows/ci.yaml#L9-L14) for avian-flu

phylogenetic/clade-iib/Snakefile
jameshadfield: +1 for exploring ways to get different configfile "entrypoints" into `nextstrain run` beyond creating multiple Snakefiles. I originally did just that for avian-flu, but my solution wasn't as nice as your proposed use of `nextstrain-pathogen.yaml` which might have been why it didn't get any traction.

Even with the `--configfile` approach we'll still have subtleties for analysis directory usage such as `"color_schemes.tsv"` vs `"clade-iib/auspice_config.json"`. I guess the solution here is [better per-pathogen docs](https://github.com/nextstrain/zika/pull/90).

PR #335, opened 2025-09-01
Fix snakefmt-induced bug

corneliusroemer:
## Checklist

- [ ] Checks pass
- [ ] Update changelog

### DIFF
--- a/phylogenetic/rules/config.smk
+++ b/phylogenetic/rules/config.smk
@@ -103,8 +103,8 @@ assert len(
 ROOT_FLAG = "--root "
 if config.get("treefix_root", "").startswith(ROOT_FLAG):
     print(
-        f"Removing the flag {ROOT_FLAT! r} from config['treefix_root'] ; "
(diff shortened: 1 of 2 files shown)


phylogenetic/rules/config.smk
```diff
@@ -103,8 +103,8 @@ assert len(
 ROOT_FLAG = "--root "
 if config.get("treefix_root", "").startswith(ROOT_FLAG):
     print(
-        f"Removing the flag {ROOT_FLAT! r} from config['treefix_root'] ; "
-        f"consider updating the config param in the config file.",
+        f"Removing the flag {ROOT_FLAT} from config['treefix_root'] ; "
```
victorlin, line 106: Not a huge deal, but isn't this and the direct commit b36dfce0e534236a828067be655258fba6ed7df5 more of a workaround than a fix? It would be good to restore `!r` after the snakefmt bug is fixed.

PR #344, opened 2025-11-13
ingest/extract_open_data: Do not compress intermediate files

joverlee521:

Use uncompressed outputs so that the uploaded files are no longer double compressed.¹

Once `upload-to-s3` supports compressed local files, we can go back to using compressed intermediate files.

## Related issue(s)

Related to https://github.com/nextstrain/mpox/issues/343

## Checklist

- [ ] Checks pass
- [ ] Update changelog

### DIFF
--- a/ingest/Snakefile
+++ b/ingest/Snakefile
@@ -48,8 +48,8 @@ rule extract_open_data:
         metadata="results/metadata.tsv",
         sequences="results/sequences.fasta",
     output:
-        metadata="results/metadata_open.tsv.zst",
-        sequences="results/sequences_open.fasta.zst",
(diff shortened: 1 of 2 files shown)


corneliusroemer:
Thanks and sorry for this. How did you notice the double compression? It hasn't caused any issues in any of our CI runs as far as I can tell?

joverlee521:
> the automated phylo workflows start from the *_with_restricted files, so nothing flagged this issue.

It would be good for us to use both the OPEN and RESTRICTED files _and_ to not host duplicate OPEN records on S3. Updating ingest to upload completely separate files for OPEN and RESTRICTED records in https://github.com/nextstrain/mpox/pull/347

PR #357, opened 2026-02-03
config: Note Pathoplexus data use terms

victorlin:
Without a note, users would have to check the Data Use Terms columns for compliance, which might not be an obvious step.


jameshadfield (approved):

victorlin:
Closing in favor of bigger changes described in https://github.com/nextstrain/public/issues/36

PR #376, opened 2026-09-16
Update scheduled workflows

victorlin:
The last commit directly addresses this morning's [failed scheduled run](https://github.com/nextstrain/mpox/actions/runs/35130697915). Everything else is additional improvements that I came up with while poking around.

## Checklist

- [x] Trial runs for [mpxv](https://github.com/nextstrain/mpox/actions/runs/35161125918), [hmpxv1](https://github.com/nextstrain/mpox/actions/runs/35161210812) on GitHub-hosted runners succeed
- [x] Checks pass
- [x] ~Update changelog~ no user-facing changes
- [ ] Post-merge: next ingest run is successful


.github/workflows/rebuild-clade-i.yaml
```diff
@@ -44,15 +44,21 @@ jobs:
     with:
       # We can migrate to AWS Batch when/if we need to for more resources,
       # but at the time of writing the clade-I build is small & quick
-      runtime: docker
+      runtime: aws-batch
```
joverlee521, line 47: _non-blocking_

When the clade-I build was first added in 2024, there were <200 sequences. ~2 years later there are now [3,800 records on PPX](https://pathoplexus.org/mpox/search?clade=I&clade=Ia&clade=Ib). I wonder if we should add subsampling instead of just taking all clade-I 

https://github.com/nextstrain/mpox/blob/1a0ff7c317ce5875654d440cf52ee140a6f7b96b/phylogenetic/defaults/clade-i/config.yaml#L30-L33

If subsampled, this build may still be able to use the GH Action runner with the docker runtime?

.github/workflows/fetch-and-ingest.yaml
```diff
@@ -43,17 +43,13 @@ jobs:
     uses: nextstrain/.github/.github/workflows/pathogen-repo-build.yaml@master
     secrets: inherit
     with:
-      runtime: aws-batch
```
joverlee521, line 46: I _think_ the ingest workflow was limited by the 14GB SSD on Github-hosted runners not the 16GB memory. However, that was before we switched to PPX where we can directly fetch the [zstd compressed NDJSON](https://github.com/nextstrain/mpox/blob/1a0ff7c317ce5875654d440cf52ee140a6f7b96b/ingest/rules/fetch_from_ppx.smk#L23). 

After downloading the last [ingest run from AWS Batch](https://github.com/nextstrain/mpox/actions/runs/35127655001), the ingest directory is only 8.8G so I think we're fine for now.

joverlee521 (approved):
Thanks for the series of improvements here!

.github/workflows/rebuild-clade-i.yaml
victorlin: Good point. Started a [Slack thread](https://bedfordlab.slack.com/archives/C03G8HQEV18/p1790035045585049).

